PR #10, opened 2015-03-02
Add ability to always associate to OS

sathieu:
with associate=always

PR #11, opened 2015-03-12
Handling for multibranch templates repo

GregSutcliffe:
@ekohl @domcleal in the context of the discussion for branching community-templates, here's a patch for handling that. I'll add to the discussion on the mailing list for the scenarios.

### DIFF
--- a/lib/templates.rake
+++ b/lib/templates.rake
@@ -1,6 +1,7 @@
 require 'fileutils'
 require 'yaml'
 require 'diffy'
+require 'git'
 
 class NoKindError < Exception ; end
 
@@ -103,6 +104,17 @@ def update_snippet
   { :diff => diff, :status => status, :result => result }
 end
 
+def get_default_branch(repo)
+  branch_names = repo.branches.map(&:name).uniq
+
+  # Always use develop on Foreman-nightly, if present, or else relevant stable branch
+  target = SETTINGS[:version].tag == 'develop' ? 'develop' : "#{SETTINGS[:version].short}-stable"
+  return target if branch_names.include?(target)
+
+  # stay on default branch as fallback
+  return nil
+end
+
 desc <<-END_DESC
 Synchronize templates from a git repo
 END_DESC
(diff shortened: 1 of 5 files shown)


lib/templates.rake
```diff
@@ -103,6 +107,23 @@ def update_snippet
...
+def get_default_branch
+  # Always use develop on Foreman-nightly
+  return "-b develop" if Foreman::Version.new.tag == "develop"
+
+  url = URI.parse("https://api.github.com/repos/theforeman/community-templates/branches")
+  req = Net::HTTP.new(url.host, url.port)
+  req.use_ssl = true
+  res = req.get(url.path)
+  repodata = JSON.parse(res.body)
+  version = Foreman::Version.new.short
```
domcleal: `SETTINGS[:version]` rather than creating a new object.

lib/templates.rake
```diff
@@ -103,6 +107,23 @@ def update_snippet
   { :diff => diff, :status => status, :result => result }
 end
 
+def get_default_branch
+  # Always use develop on Foreman-nightly
+  return "-b develop" if Foreman::Version.new.tag == "develop"
+
+  url = URI.parse("https://api.github.com/repos/theforeman/community-templates/branches")
```
ekohl: Can this somehow be deduced from the git repo? That would also allow users with internal git repos to use multiple branches.

lib/templates.rake
GregSutcliffe: Oops, I knew there was a better way. Thanks.

GregSutcliffe: Is there an equivalent for :tag so I can check for "develop" or do I need to do some string parsing?

domcleal: `SETTINGS[:version]` is a Foreman::Version object, so `SETTINGS[:version].tag` or .short, notag etc will work.

lib/templates.rake
GregSutcliffe: At the moment we do the branch as part of the clone command (line 156). We could clone first I guess and do some shell logic, or maybe find a git gem to parse it. Certainly possible.

GregSutcliffe: When you say "allow users with internal git repos to use multiple branches", how do you mean? Users can already specify a custom branch anyway, this is just to handle defaults for the templates repo transition

domcleal: There are ruby/git bindings, by the way.  We use them in https://github.com/theforeman/theforeman.org/blob/gh-pages/scripts/committers.rb, which I notice, er, that you wrote :smile:

ekohl: This may be an obscure use case, but never underestimate what create solutions users can come up with. An organization could have a production setup and a separate lab setup to test any changes. They have provisioning templates stored in their own git repository to sync them between the two setups. Now the lab has a branch lab with the latest changes while production uses production.

Now that I think about it, it's a bigger problem. If they have a custom git repo, the default branch may exist in our github but not in theirs. I don't think we can assume a repo layout for all our users.

ekohl: I now see you already noticed this could be a problem. Still, could be nice to detect the branches for any type of repository, not just ours.

GregSutcliffe: Makes sense. I think it'll be fine - it's better to check out the branch on the commandline for repos containing submodules (otherwise switching branch can get messy) but we've no expectation of using submodules in a templates repo ;)

@domcleal hah, that was a while ago. I may steal some of my own code :P

lib/templates.rake
GregSutcliffe: Perfect, thanks. Will update.

GregSutcliffe: Fixed.

GregSutcliffe:
@ekohl @domcleal updated. This is much nicer, I think (as it uses Git.clone instead of `git ...`).

ekohl:
Much cleaner :+1:

mmoll:
Looking good and working for me, so :+1:

GregSutcliffe:
Merged thanks guys. I'll see if I can quickly add some tests to #10 and then we can get a release out

PR #13, opened 2015-04-29
Update link to foreman plugin installation

andskli:
I think this is a better link to the plugin installation documentation, as the old link points to what seems to be an outdated page that links forward to the proposed link.


domcleal:
Thanks @andskli!  (Also I'm just fixing 'latest' to point to 1.8.)

PR #15, opened 2015-06-02
Only remove temp dir if it still exists

domcleal:
git cloning causes the dir to be removed, so an error raised during the
clone leaves the dir missing when we reach the `ensure` handler.


ekohl:
:+1:

PR #20, opened 2015-12-01
Fixes #12640 - set "Community " prefix by default

brandonweeks:


GregSutcliffe:
Wow, looks like this got accidentally trashed all the way back in fef01979d508321c99b78a31b107aff31be26ec3 - good catch!

GregSutcliffe:
Merged as ae16a1bab4460033b26d67bbd6c797bdadb629ad thanks!

PR #21, opened 2015-12-29
make templates.rake Rails 4 compatible

mmoll:
Still works for me with Foreman 1.10. We need a 2.0.3 release then.

### DIFF
--- a/lib/templates.rake
+++ b/lib/templates.rake
@@ -32,7 +32,7 @@ def update_template
     raise NoKindError
   end
 
-  db_template = ProvisioningTemplate.find_or_initialize_by_name(@name)
+  db_template = ProvisioningTemplate.where(name: @name).first_or_initialize
   data = {
     :template         => @text,
     :snippet          => false,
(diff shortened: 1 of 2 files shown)


lib/templates.rake
```diff
@@ -32,7 +32,7 @@ def update_template
     raise NoKindError
   end
 
-  db_template = ProvisioningTemplate.find_or_initialize_by_name(@name)
+  db_template = ProvisioningTemplate.where(name: @name).first_or_initialize
```
domcleal: Nitpick, usually we use hash "rockets" style, `where(:name => @name)`

domcleal:
Tests pass now, so that's good.  :+1:

GregSutcliffe:
Merged as 92533afd0139261950199811ed0e5626c028a8a5, fixed the nitpick (hope you don't mind @mmoll :P). Thanks!

mmoll:
@GregSutcliffe not at all :)

PR #22, opened 2016-02-24
Support of group/environment association

quadriq:
Hi, 
I extended the script to support association by hostgroup/environemnt, using tempalte metadata

The metadata should be in format:

```
groups:
- test1
- test1/test2
- test3,production
```

so its a list of groups-titles, means it takes a "title" value, which is everything, including parent groups, separated by a slash.

and if an environment should be also included, then it should be append to the group-name, separated by a ","-comma (no spaces after comma)

I am new to RUBY, so my code may be not very good.. 

what do you think?

### DIFF
--- a/lib/templates.rake
+++ b/lib/templates.rake
@@ -9,6 +9,10 @@ def db_oses
   @db_oses || Operatingsystem.all
 end
 
+def db_groups(metagroups)
+  @db_groups = Hostgroup.where(title: metagroups)
+end
+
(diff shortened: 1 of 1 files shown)


ares:
[test]

ares:
Thanks for the patch. Would you mind rebasing it to the latest version of the plugin? I could then do the review and help with the code if needed.

ares:
Closing due to the inactivity

Issue #24, opened 2016-06-28
Works with partition table snippets?


cacack:
We're using Foreman 1.10.3 and have a number of partition table snippets.  I'd like to use this plugin to migrate to a git repository, but the plugin doesn't appear to support using snippets in partition tables.  Code leads me to believe snippets will be imported as provisioning templates.  Is this correct?

GregSutcliffe:
It shouldn't matter how you structure it. PTable templates should be correctly imported (see https://github.com/theforeman/foreman_templates/blob/master/lib/templates.rake#L81-L106 for the handler on ptable entries) and snippets (which are a different thing - they are only text with no associations or types) will be imported using this 9https://github.com/theforeman/foreman_templates/blob/master/lib/templates.rake#L108-L126)

cacack:
Thanks for the quick response!

In Foreman, we have some partition table templates with snippet checked and some not checked.  But it appears that doesn't matter to Foreman as you can use the snippet macro to source in any template of like type, not just ones marked as snippet.  I just happened to sample the ones that were marked as snippets leading to my assumption.

cacack:
I did a little more digging and discovered that Foreman supports snippets that are stored with [provision templates](https://theforeman.org/api/1.12/apidoc/v2/provisioning_templates/create.html) AND [partition tables](https://theforeman.org/api/1.12/apidoc/v2/ptables/create.html) by using the snippet param.  Foreman doesn't treat snippets as a separate [kind type](https://theforeman.org/api/1.12/apidoc/v2/template_kinds/index.html).

Can the plugin be modified to mirror this behavior?

I think the template metadata would need to support a snippet flag when the type is provision or ptable. Something like:
```
<%#
kind: ptable
snippet: true
name: my_partition_table
...
%>
```

GregSutcliffe:
Fixed in 11ee3cb79d00dfd82b44d8ec70c87772853b1715

PR #25, opened 2016-06-30
Support metadata snippet flag.

cacack:
This adds support of importing snippets as either provisioning or ptable as discussed in theforeman/foreman_templates#24.  It does so by adding a boolean to the metadata named snippet and defaults to false so backwards compatibility is maintained.

### DIFF
--- a/lib/templates.rake
+++ b/lib/templates.rake
@@ -50,7 +50,7 @@ def update_template
   db_template = ProvisioningTemplate.where(:name => @name).first_or_initialize
   data = {
     :template         => @text,
-    :snippet          => false,
+    :snippet          => @snippet,
     :template_kind_id => kind.id
   }
   string = db_template.new_record? ? "Created" : "Updated"
@@ -80,7 +80,10 @@ end
 
 def update_ptable
   db_ptable = Ptable.where(:name => @name).first_or_initialize
-  data = { :layout => @text }
+  data = {
+    :layout  => @text,
+    :snippet => @snippet
+  }
   string = db_ptable.new_record? ? "Created" : "Updated"
 
   oses = map_oses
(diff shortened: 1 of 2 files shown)


stbenjam:
[test]

stbenjam:
[test]

cacack:
I thought I fixed it.  Maybe someone can look it over?

GregSutcliffe:
It's slightly subtle. The tests try to check each separate method does it's own job correctly, so *only* `udpate_template` is called in those two tests. However, you're setting `@snippet` in the main body of the rake task - in the tests, that isn't getting called. As such, @snippet is undefined and the DB throws an error.

You probably just need to add `@snippet = false` after @associate. If you have the chance, some tests for update_ptable would be good too, I never got around to that - but don't sweat it, I'm just about to rewrite this based on the app/service stuff in the UI branch :P

cacack:
@stbenjam Pinging to get tests ran...

stbenjam:
[test]

GregSutcliffe:
Thanks @cacack! These look like temporary failures, re-running [test]

mmoll:
this needs a serious rebase now ;)

GregSutcliffe:
@mmoll well spotted :)

@cacack My bad for not getting back to this. It should be pretty easy to fix up though - the code hasn't actually changed much, only moved around. I'll see if I can fix it up for you, since I really should have dealt with it by now :P

GregSutcliffe:
@cacack @mmoll I think https://github.com/GregSutcliffe/foreman_templates/commit/be1882c5698c496dcaa64e3ad09c2a9f035f3caf should cover it, yes?

cacack:
Thanks @GregSutcliffe for moving this forward!

GregSutcliffe:
@cacack No worries. If you can look it over and maybe test it solves your usecase, that'd be cool

GregSutcliffe:
Fixed via 11ee3cb - thanks @cacack for the initial implementation!

PR #26, opened 2016-07-21
Hook rake tests on test:plugins

lzap:
See https://github.com/theforeman/foreman/pull/3666


ares:
setting WoC until core PR is finished

ares:
this needs rebase, but core PR did not move any further, it seems some tests are red in there

PR #27, opened 2016-08-23
Move the bulk of the code to a service class

GregSutcliffe:
Pretty huge rewrite here...

Big stuff:

* 99% of the Rake task is now in /app - there's a service class (ForemanTemplates::TemplateImporter) you can call from the console (eg `ForemanTemplates::TemplateImporter.new(verbose: true).import!`)
* Much of the code is moved to extensions on `Ptable` and `ProvisioningTemplate` - this makes it possible to call `ProvisioningTemplate.import!`.
* The main `case` block supports detecting `metadata['model']` and using it to call `import!` directly - this means other plugins can add their own import methods that ForemanTemplates can use (eg REX)

Smaller stuff:
* Tests rewritten to match new code - probably some missing cases, working on getting a code coverage report.
* Updated the lib/ folder to match current plugin_template standards
* Rubocop linting (mostly) applied.

@mmoll @stbenjam care to take a look? I'd love some eyes on this, and some human testing, before I merge (and version bump to 3.0). The new `model` option for templates will be handy, but obviously this has to get out in the wild before we can start using it.


app/services/foreman_templates/template_importer.rb
```diff
@@ -0,0 +1,140 @@
...
+    end
+
+    def parse_metadata(text)
+      # Pull out the first erb comment only - /m is for a multiline regex
+      extracted = text.match(/<%\#(.+?).-?%>/m)
+      extracted.nil? ? {} : YAML.load(extracted[1])
+    end
+
+    def update_job_template(name, text)
+      return {
```
GregSutcliffe: @stbenjam would you support a deprecation warning here? We can ask people to update their templates to have `metadata['model'] = 'job_template'` before 4.0?

GregSutcliffe:
I'm sure there's more refactoring and/or better ways to write some of this, but hopefully it's a step in the right direction...

app/services/foreman_templates/template_importer.rb
stbenjam: Sure, I think that's ok

GregSutcliffe:
Tests updated to match new implementation. I think this is good to go. More eyes definitely welcome!

GregSutcliffe:
Tests do pass - http://ci.theforeman.org/job/test_plugin_pull_request/2874/

PR #29, opened 2016-08-31
WIP - RFC 'purge'

matonb:
Add rake task purge templates.

Work in progress, please  comment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,32 @@ get written to a file in `/tmp`.
   * `:old` and `:new` (in which case this plugin will calculate the diff)
 * :result` (text, may be nil).
 
+=======
+### Purge
+
+    foreman-rake templates:purge
+
+This will remove all templates with the prefix "Community "
+#### Rake options
+
+* prefix    => The string all templates to be purged should begin with [Community ]
+* negate    => Negate the search [false]
+* verbose   => Print extra information during the run [false]
+
+#### Examples
+
+Just purge all the templates the begin with 'Community '
+
+    foreman-rake templates:purge
+
+Purge all templates that begin with 'Oops '
+
+    foreman-rake templates:purge prefix='Oops '
+
+Purge all templates that do not begin with 'Community '
+
+    foreman-rake templates:purge negate=true
+
 ## TODO
 
 * Add a button to the UI with Deface to run the rake task
--- a/app/services/foreman_templates/template_importer.rb
+++ b/app/services/foreman_templates/template_importer.rb
@@ -13,6 +13,7 @@ module ForemanTemplates
       @dirname   = args[:dirname] || '/'
       @filter    = args[:filter] || nil
       @repo      = args[:repo] || 'https://github.com/theforeman/community-templates.git'
+      @negate    = args[:negate] || false
       @branch    = args[:branch] || false
 
       # Rake hands off strings, not booleans, and "false" is true...
@@ -154,5 +155,14 @@ module ForemanTemplates
       result = "  #{c_or_u} Template #{id_string}:#{name}"
       { :diff => diff, :status => template.save, :result => result }
     end
+
+    def purge!
+      clause = "name #{@negate?'NOT ':''}LIKE ?"
+      ProvisioningTemplate.where(clause, "#{@prefix}%").each do |template|
+        puts template if @verbose
+        template.destroy
+      end
+    end # :purge
+
   end
 end
--- a/lib/tasks/foreman_templates_tasks.rake
+++ b/lib/tasks/foreman_templates_tasks.rake
@@ -23,6 +23,19 @@ namespace :templates do
 
     puts results.join("\n")
   end
+
+  desc 'Purge unwanted templates from foreman'
+  task :purge => :environment do
+    ForemanTemplates::TemplateImporter.new({
+      #* negate  => negate query [false]
+      #* prefix  => The string all templates to purge should ( or not ) begin with [Community ]
+      #* verbose => Print extra information during the run [false]
+      negate:  ENV['negate'],
+      prefix:  ENV['prefix'],
+      verbose: ENV['verbose'],
+    }).purge!
+  end
+
 end
 
 # Tests
--- a/test/unit/template_importer_test.rb
+++ b/test/unit/template_importer_test.rb
@@ -142,5 +142,26 @@ module ForemanTemplates
       assert_equal File.read(get_template('ptable1.erb')), pt.layout
       assert_equal 'Debian', pt.os_family
     end
+
+    test 'purge! removes expected templates' do
+      # Create more than one template as we expect to delete all 'FooBar '
+      # templates.
+      FactoryGirl.create(:provisioning_template, :name => 'FooBar delete_me')
+      FactoryGirl.create(:provisioning_template, :name => 'FooBar delete_me_too')
+      FactoryGirl.create(:provisioning_template, :name => 'keep_me')
+      count = ProvisioningTemplate.all.size
+      @importer.purge!
+      refute ProvisioningTemplate.find_by_name('FooBar %').present?
+      assert ProvisioningTemplate.find_by_name('keep_me').present?
+      assert ProvisioningTemplate.all.size, count - 2
+    end # 'purge! removes expected template'
+
+    test 'purge! negated purge leaves expected templates' do
+      FactoryGirl.create(:provisioning_template, :name => 'FooBar keep_me')
+      @importer = importer(:negate => true)
+      @importer.purge!
+      assert ProvisioningTemplate.find_by_name('FooBar keep_me').present?
+      assert ProvisioningTemplate.all.size, 1
+    end # 'purge! removes expected template'
   end
 end


GregSutcliffe:
I like where this is going :)

You may want to rebase off #27 which I'll be merging in a day or two, and then we'll probably want some tests that ensure the right things are deleted and other templates are not. Let me know if you need help with that.

matonb:
A bit of help would be appreciated, clone/checkout/pull/push is about the extend of my everyday git usage :)

I'm not even sure how to checkout '#27'...

Cheers

GregSutcliffe:
The absolute easiest is to wait a day or two until #27 lands in master :p

Otherwise, you would (a) add my fork as a remote (via `git remote add`, and call it e.g. `greg`), (b) run git fetch to makes sure you have data for my fork, and then (c) run `git rebase greg/serviceclass` which will apply your patch on top of my changes. You know it's right if you look in `git log` and see my commits there :)

matonb:
I'll give it a go, I really really ought to know more about git ;)

matonb:
@GregSutcliffe I think that got it

GregSutcliffe:
Yep, that got it, nice one. I'm assuming the task still works for you (since it doesn't depend on my changes) but it does mean you should be able to pretty easily write a couple of tests. I'll add a few inline comments...

matonb:
Great, thanks for that.  I'm all up for tests.

I haven't had chance to test it yet, on the agenda for tomorrow morning....

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -0,0 +1,82 @@
...
+    @verbose = ( ENV['verbose'] and ENV['verbose'] != 'false' ) ? true : false
+    match    = ENV['not'] ? 'NOT LIKE' : 'LIKE'
+    kind     = ENV['kind'] ? ENV['kind'] : "template_kind #{match} '%'"
+    prefix   = ENV['prefix'] ? ENV['prefix'] : 'Community '
+    filter   = ENV['filter'] ? ENV['filter'] : "name #{match} '#{prefix}%'"
+
+    #clause = [filter, kind].join(" AND ")
+    ProvisioningTemplate.where(filter).each do |template|
+      puts template if @verbose
+      template.destroy
```
GregSutcliffe: This obviously works, but the point of #27 is to move the executable actions out of the Rakefile and into a service class. This is better because when I (eventually) write a UI for the plugin, I don't want to be calling Rake tasks from the UI - rather, both the UI and the Rake task should use a shared codebase

Could you move this to a trivial method in the Importer class (`app/services/foreman_templates/template_importer.rb`) so that it's available there? It should be a case of copying whats on line 14-22 here and calling `.purge!` or whatever you call the method

test/unit/template_importer_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      assert_equal 'provision', hash['kind']
+      assert_equal 'Test Data', hash['name']
+      assert_equal 5, hash['oses'].size
+
+      text = File.read(get_template('metadata2.erb'))
+      hash = @importer.parse_metadata(text)
+      assert_equal 'ProvisioningTemplate', hash['model']
+    end
+
+    test 'import! loads templates into db' do
```
GregSutcliffe: Once you've got `purge!` set up in the importer class, we'll want to add a test similar to this one with 2 assertions. You'll want to create a couple of templates (something like `FactoryGirl.create(:provisioning_template, :name => 'blah')` should do), one with a prefix, one without. You can the call `.purge!` and assert that the correct one is gone and the other is still present.

GregSutcliffe: Again, if you need a hand with that, let me know.

matonb: Sure first rebase was just to shoe horn purge in.

Thanks for the pointers, I'll try and squeeze it in tomorrow.

GregSutcliffe: No worries. I have one more test to add to #27 and then I can merge it, which will make your fork simple again ;)

matonb: Hi Greg,

  Made the changes, not pushed yet as I can't run the tests and I'm don't
know what to do about this error:

bundle install
bundle exec rake

~/.gem/ruby/gems/bundler-1.11.2/lib/bundler/rubygems_integration.rb:304:in
`block in replace_gem': rake is not part of the bundle. Add it to Gemfile.
(Gem::LoadError)
        from /home/vagrant/bin/rake:22:in `<main>'


Dev box:
OEL 7.2
Ruby ruby 2.0.0p598 (2014-11-13) [x86_64-linux]


Any pointers appreciated,

matonb:
*cough* just read the message again....

app/services/foreman_templates/template_importer.rb
```diff
@@ -131,5 +131,28 @@ def update_job_template(name, text)
...
+
+    def purge!
+      # Available options:
+      #* verbose => Print extra information during the run [false]
+      #* not     => negate prefix / filter
+      #* kind    => Template kind
+      #* prefix  => The string all imported templates should begin with [Community ]
+      #* filter  => Import names matching this regex (case-insensitive; snippets are not filtered)
+
+      @verbose = ( ENV['verbose'] and ENV['verbose'] != 'false' ) ? true : false
```
GregSutcliffe: You can't use ENV in app/services, it won't exist. If you look at how the Rake task calls `import!` it passes the appropriate ENV variables in as options. Over-complicated I know, but it's for future stuff, I promise ;)

test/unit/template_importer_test.rb
```diff
@@ -92,5 +92,15 @@ def importer(opts = {})
       assert_equal File.read(get_template('ptable1.erb')), pt.layout
       assert_equal 'Debian', pt.os_family
     end
+
+    test 'purge! removes expected template' do
+      FactoryGirl.create(:provisioning_template, :name => 'Community delete_me')
+      FactoryGirl.create(:provisioning_template, :name => 'keep_me')
+      @importer.purge!
+      ct = ProvisioningTemplate.find_by_name('delete_me')
+      assert not ct.present?
```
GregSutcliffe: You can use `refute` here, looks good otherwise :+1:

GregSutcliffe:
Gah just added comments to the commits instead of the PR, let me move that...

app/services/foreman_templates/template_importer.rb
```diff
@@ -0,0 +1,158 @@
...
+      match    = ENV['not'] ? 'NOT LIKE' : 'LIKE'
```
GregSutcliffe: you don't seem to use `match` at all?

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -0,0 +1,61 @@
...
+      repo:      ENV['repo'],
+      branch:    ENV['branch'],
+      prefix:    ENV['prefix'],
+      dirname:   ENV['dirname'],
+      filter:    ENV['filter'],
+      associate: ENV['associate'],
+    }).import!
+  end
+
+end
```
GregSutcliffe: You still need most of the `purge!` code here - as mentioned above, it's literally just 151-154 of the app class that needs to stay where it is :)

app/services/foreman_templates/template_importer.rb
matonb: Yeah I noticed the stuff in import!, not sure why I didn't update everything ;)

test/unit/template_importer_test.rb
matonb: Cool, I was looking at the a assertion stuff for a negative test, obviously missed refute, thanks for that.

app/services/foreman_templates/template_importer.rb
matonb: I'll have a look, I figured I'd use it to negate a search might not have used it yet...

lib/tasks/foreman_templates_tasks.rake
matonb: Ah -ha that would help a bit :)

matonb:
How do I run the tests, doesn't seem geared for a rake task ?

GregSutcliffe:
If you run `bundle exec rake -T` in your *Foreman* directory (not the plugin checkout) then you should see a `test:foreman_templates` task along with the usual `templates:sync` task. Run that one :)

app/services/foreman_templates/template_importer.rb
GregSutcliffe: Ah so you could wipe out everything *except* the specified filter? Nice

app/services/foreman_templates/template_importer.rb
GregSutcliffe: In fairness, I did say "move it all to the service class" which was a little misleading :P

app/services/foreman_templates/template_importer.rb
```diff
@@ -133,21 +133,17 @@ def update_job_template(name, text)
...
-      #* prefix  => The string all imported templates should begin with [Community ]
-      #* filter  => Import names matching this regex (case-insensitive; snippets are not filtered)
-
-      @verbose = ( ENV['verbose'] and ENV['verbose'] != 'false' ) ? true : false
-      match    = ENV['not'] ? 'NOT LIKE' : 'LIKE'
-      #kind     = ENV['kind'] ? ENV['kind'] : "template_kind #{match} '%'"
-      prefix   = ENV['prefix'] ? ENV['prefix'] : 'Community '
-      filter   = ENV['filter'] ? ENV['filter'] : "name #{match} '#{prefix}%'"
+#      @kind    = args[:kind] || 'something'
+      @match   = args[:negate] ? 'NOT LIKE ' : 'LIKE '
```
GregSutcliffe: I'd still leave this out if you're not going to use it later on - it's just a code-clarity thing.

GregSutcliffe:
Starting to look pretty solid! I'll test it shortly.

matonb:
Ok tests....
I've got a clones of foreman and puppet-foreman on my dev box, but that's not running foreman or puppet.
I have another server which is built by foreman/puppet that is used for testing, but as far as I can tell that's install foreman_template from an RPM package.
I could copy the code over I suppose, better still set something up to test on my dev box ?

GregSutcliffe:
You don't need puppet-foreman, that's just for the installer. If you want to get the tests running locally, you just need a dev foreman setup and the foreman_templates plugin (both from git, ideally). 

https://theforeman.org/contribute.html#Code-relatedcontributions has the details for getting set up, but essentially its (1) add templates to the Gemfile.local.rb, (2) bundle install && npm install, (3) db:migrate (4) bundle exec rake test:foreman_templates

GregSutcliffe:
If you want to some help getting the tests running in realtime, feel free to hit me up in IRC - gwmngilfen is my nick

matonb:
I got a copy of puppet-foreman to locate 'stuff', the joy of grep :-p
I'll have a look at setting up a foreman test environment, thanks for the pointers.

And thanks for the IRC offer, will take you up on that.

GregSutcliffe:
FYI #27 is merged so you can rebase my stuff out of this PR now ;)

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -0,0 +1,75 @@
...
+      repo:      ENV['repo'],
+      branch:    ENV['branch'],
+      prefix:    ENV['prefix'],
+      dirname:   ENV['dirname'],
+      filter:    ENV['filter'],
+      associate: ENV['associate'],
+    }).import!
+  end
+
+    ForemanTemplates::TemplateImporter.new({
```
GregSutcliffe: You've not started a new task here (:task purge ...) so when you put `end` after the block, it causes a syntax error

app/services/foreman_templates/template_importer.rb
```diff
@@ -0,0 +1,154 @@
...
+      result = "  #{c_or_u} Template #{id_string}:#{name}"
+      { :diff => diff, :status => template.save, :result => result }
+    end
+
+    def purge!
+#      @kind    = args[:kind] || 'something'
+      @match   = args[:negate] ? 'NOT LIKE ' : 'LIKE '
+      @prefix  = args[:prefix] || 'Community '
+      @filter  = args[:filter] || "name #{match} '#{prefix}%'"
+      @verbose = args[:verbose] || false
```
GregSutcliffe: you don't need to use `args` here - that only available in `initialize` - but `@` variables are class-wide so you can just use them right away, so long as they get set properly in `initialize`. Currently, `@match` is the only one not set up, so if you drop that, you don't need any changes in `initialize` and can just use `@prefix/filter/verbose` directly (as `import!` and `parse_files!` do)

test/unit/template_importer_test.rb
```diff
@@ -0,0 +1,106 @@
...
+      pt = Ptable.find_by_name('FooBar Test Ptable')
+      assert pt.present?
+      assert_equal File.read(get_template('ptable1.erb')), pt.layout
+      assert_equal 'Debian', pt.os_family
+    end
+
+    test 'purge! removes expected template' do
+      FactoryGirl.create(:provisioning_template, :name => 'Community delete_me')
+      FactoryGirl.create(:provisioning_template, :name => 'keep_me')
+      @importer.purge!
```
GregSutcliffe: the setup method for `@importer` (lines 16-21 in this file) don't define a `filter` only a `prefix` - so filter is `nil` and the destroy code in `app` deletes every template. You probably want to pass a filter in, like `@importer = importer(:filter => 'Community ')

test/unit/template_importer_test.rb
```diff
@@ -0,0 +1,106 @@
...
+      ct = ProvisioningTemplate.find_by_name('delete_me')
```
GregSutcliffe: This can never fail - you're creating "Community delete_me" but testing for presence of "delete_me" - you need to update the name here to include the prefix

test/unit/template_importer_test.rb
```diff
@@ -0,0 +1,106 @@
...
+      refute ct.present?
+      ct = ProvisioningTemplate.find_by_name('keep_me')
+      assert ct.present?
+    end # 'purge! removes expected template'
```
GregSutcliffe: You could also show no other templates were deleted using "ProvisioningTemplate.all.size" before and after calling `purge!` - they should differ by exactly 1, right? :)

app/services/foreman_templates/template_importer.rb
```diff
@@ -0,0 +1,154 @@
...
+
+      # TODO add filter on template 'kind'
+      # kind is an ID, need to map / match with lookup table
+      #kind     = ENV['kind'] ? ENV['kind'] : "template_kind #{match} '%'"
+      #clause = [filter, kind].join(" AND ")
+      # TODO: Make query more flexible, robust and secure from SQL injection
+      ProvisioningTemplate.where(filter).each do |template|
```
GregSutcliffe: Using `ProvisioningTemplate` here will exclude purging Ptable templates - if you want to include them all in the purge, just use the base class `Template` - this goes for the tests too (expect the FactoryGirl calls where you want to create a specific type of template)

GregSutcliffe:
Ok, tested the code and found a few issues, but otherwise it works well. I did see a small conflict when rebasing out the chagnes in #27 (so that we get back to just your commits in the branch), let me know if that catches you out (I did `rebase -i` and removed the offending commit about simplecov, for reference)

matonb:
Hi Greg which IRC channel are you on ?

ekohl:
We are on #theforeman-dev which lives on Freenode. His nick is gwmngilfen, or gw<tab> as I remember it.

README.md
```diff
@@ -122,3 +149,4 @@ GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
```
GregSutcliffe: nitpick: empty line added

app/services/foreman_templates/template_importer.rb
```diff
@@ -154,5 +155,19 @@ def update_job_template(name, text)
...
       { :diff => diff, :status => template.save, :result => result }
     end
+
+    def purge!
+      # TODO add filter on template 'kind'
+      # kind is an ID, need to map / match with lookup table
+      #clause = [filter, kind].join(" AND ")
+      # TODO: Make query more flexible, robust and secure from SQL injection
+      match  = @negate ? 'NOT LIKE' : 'LIKE'
+      clause = @filter || "name #{match} '#{@prefix}%'"
```
GregSutcliffe: Are we expecting the user to supply `@filter` with a value like `name LIKE '%foobar%'`? This looks like it needs some SQL handling if so.

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -23,6 +23,23 @@ namespace :templates do
...
+  desc 'Purge unwanted templates from foreman'
+  task :purge => :environment do
+    ForemanTemplates::TemplateImporter.new({
+      #* filter  => Import names matching this regex (case-insensitive; snippets are not filtered)
+      #* kind    => Template kind
+      #* not     => negate prefix / filter
+      #* prefix  => The string all imported templates should begin with [Community ]
+      #* verbose => Print extra information during the run [false]
+      filter:  ENV['filter'],
+#      kind:    ENV['kind'],
```
GregSutcliffe: nitpick: no need to leave commented code in

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -23,6 +23,23 @@ namespace :templates do
 
     puts results.join("\n")
   end
+
+  desc 'Purge unwanted templates from foreman'
+  task :purge => :environment do
+    ForemanTemplates::TemplateImporter.new({
+      #* filter  => Import names matching this regex (case-insensitive; snippets are not filtered)
```
GregSutcliffe: This description seems wrong, this task doesn't import

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -23,6 +23,23 @@ namespace :templates do
...
+      #* kind    => Template kind
+      #* not     => negate prefix / filter
+      #* prefix  => The string all imported templates should begin with [Community ]
```
GregSutcliffe: Likewise, this isn't an import task

README.md
```diff
@@ -102,6 +102,33 @@ get written to a file in `/tmp`.
...
 
+=======
+### Purge
+
+    foreman-rake templates:purge
+
+This will remove all templates with the prefix "Community "
+#### Rake options
+
+* kind      => TODO
```
GregSutcliffe: Are you planning to add this type of matching in this PR or later?

README.md
```diff
@@ -102,6 +102,33 @@ get written to a file in `/tmp`.
...
+
+Just purge all the templates the begin with 'Community'
+
+    foreman-rake templates:purge
+
+Purge all templates that begin with 'Oops'
+
+    foreman-rake templates:purge prefix="Oops "
+
+Purge all templates that begin with "Community "
```
GregSutcliffe: This is the same description as line 120 - I think you mean "don't begin with" :)

app/services/foreman_templates/template_importer.rb
GregSutcliffe: Actually, you don't list filter in the task arguments at all (see README) - do we need both? is prefix enough?

GregSutcliffe:
Looking pretty good! Just a few nitpicks and a question to solve around the `where` logic.

README.md
matonb: This'll have to be later, I'll take it out for now

README.md
matonb: Just checking :-p

app/services/foreman_templates/template_importer.rb
matonb: You're quite right, it's totally vulnerable to SQL injection.
It was more for me to play with.

I'm not familiar enough with ActiveRecord to do this now, I'll take it out and limit the search to basically 'prefix' with the option to negate the action,

README.md
GregSutcliffe: I thought as much, feel free to open an issue to track it

app/services/foreman_templates/template_importer.rb
GregSutcliffe: I think thats sufficient for the vast majority of usecases and a ton safer ;)

There's one more safety optimisation you can do. Try `.where('name LIKE ?', some_variable)` for some extra Rails-enabled security.

GregSutcliffe: Obviously you can refactor this how you like, the `?` in the string is the magic bit.

test/unit/template_importer_test.rb
```diff
@@ -142,5 +142,16 @@ def self.import!(name, text, _metadata)
       assert_equal File.read(get_template('ptable1.erb')), pt.layout
       assert_equal 'Debian', pt.os_family
     end
+
+    test 'purge! removes expected templates' do
+      FactoryGirl.create(:provisioning_template, :name => 'FooBar delete_me')
+      FactoryGirl.create(:provisioning_template, :name => 'FooBar delete_me_too')
```
GregSutcliffe: What value does creating a second template provide? Does it prove anything extar over creating just one?

test/unit/template_importer_test.rb
```diff
@@ -142,5 +142,16 @@ def self.import!(name, text, _metadata)
...
+      FactoryGirl.create(:provisioning_template, :name => 'keep_me')
+      count = ProvisioningTemplate.all.size
+      @importer.purge!
+      refute ProvisioningTemplate.find_by_name('FooBar %').present?
+      assert ProvisioningTemplate.find_by_name('keep_me').present?
+      assert ProvisioningTemplate.all.size == count - 2
```
GregSutcliffe: I'd write this as `assert_equal ProvisioningTemplate.all.size, count - 2`

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -23,6 +23,19 @@ namespace :templates do
 
     puts results.join("\n")
   end
+
+  desc 'Purge unwanted templates from foreman'
+  task :purge => :environment do
+    ForemanTemplates::TemplateImporter.new({
+      #* negate  => negate query
```
GregSutcliffe: nitpick: no default listed, just `[false]` should be fine

README.md
```diff
@@ -102,6 +102,32 @@ get written to a file in `/tmp`.
...
+    foreman-rake templates:purge
+
+This will remove all templates with the prefix "Community "
+#### Rake options
+
+* prefix    => The string all templates to be purged should begin with [Community ]
+* negate    => Negate the search [false]
+* verbose   => Print extra information during the run [false]
+
+## Examples
```
GregSutcliffe: nitpick: this should probably be 4 `#`s

test/unit/template_importer_test.rb
```diff
@@ -142,5 +142,16 @@ def self.import!(name, text, _metadata)
...
+    end # 'purge! removes expected template'
```
GregSutcliffe: You should probably add a test that shows the `negate` working, so that we don't break that function in later patches. Should be pretty easy, but do it as a new `test` block rather than more assertions in this one.

matonb: i was going to ask if the changes are only effective for the test run ( not persisting on consequent executions ).

README.md
matonb: ditto

lib/tasks/foreman_templates_tasks.rake
matonb: Consider it done

test/unit/template_importer_test.rb
matonb: Yup, no problem, my roobness showing through again :)

test/unit/template_importer_test.rb
matonb: I added the second one to prove that all templates beginning with.. get deleted, not really necessary though.

test/unit/template_importer_test.rb
matonb: Actually I did try a negate, but wasn't sure how to pass the parameter to @importer.purge!

GregSutcliffe: You almost had it, just one character out :)

See https://github.com/theforeman/foreman_templates/blob/master/test/unit/template_importer_test.rb#L81 for an example, but it's expecting an options hash, so you want `:negate => true`

test/unit/template_importer_test.rb
GregSutcliffe: Ah, that it's not just a singleton delete? I suppose thats fair. Up to you then, but add a comment explaning that if you want to leave it in.

matonb: Comment coming up

GregSutcliffe:
Merged as c8eef45be573db14c3360421c8fe34147917ba66 - thanks for your patience @matonb!

PR #33, opened 2016-10-25
Fixes #16983 - Move options from importer to settings

xprazak2:


app/services/foreman_templates/template_importer.rb
```diff
@@ -163,5 +155,13 @@ def purge!
         template.destroy
       end
     end # :purge
+
+    private
+
+    def assign_attributes(args = {})
+      %w(verbose associate prefix dirname filter repo negate branch).each do |attribute|
+        instance_variable_set("@#{attribute}", args[attribute.to_sym] || Setting["templates_#{attribute}".to_sym])
```
GregSutcliffe: Some tests would be good. I think the existing tests will all hit the left-hand-side of this `||`, so some tests using `Setting` and some using both (showing that override works) would be great.

app/models/setting/templates.rb
```diff
@@ -0,0 +1,29 @@
...
+    self.transaction do
+      [
+        self.set('templates_verbose', N_('Choose verbosity for Rake task importing templates'), false, N_('Verbosity')),
+        self.set('templates_associate', N_('Associate templates to OS ("always", "new", "never")'), "new", N_('Associate')),
+        self.set('templates_prefix', N_('The string all imported templates should begin with'), "Community ", N_('Prefix')),
+        self.set('templates_dirname', N_('The directory within the git tree containing the templates'), '/', N_('Dirname')),
+        self.set('templates_filter', N_('Import names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
+        self.set('templates_repo', N_('Default Git repo to sync from'), 'https://github.com/theforeman/community-templates.git', N_('Repo')),
+        self.set('templates_negate', N_('Negate the search'), false, N_('Negate')),
+        self.set('templates_branch', N_('Default branch in Git repo'), nil, N_('Branch'))
```
GregSutcliffe: This used to default to `false` - I don't think it makes a difference, just wanted to check that this is intentional?

xprazak2: Yes, it is intentional. If you keep `false`, the in-place form for editing settings assumes the value should be boolean and offers Yes/No alternatives instead of input field.

app/services/foreman_templates/template_importer.rb
xprazak2: I'll be back with tests...

app/models/setting/templates.rb
GregSutcliffe: Ah of course, that makes sense. Tests have passed and I think that will cover this change, so that's fine. Thanks!

stbenjam:
You can extend the Settings helper and provide the values to select from instead of the validation methods:

https://github.com/sean797/katello/commit/e428c673b0b4474ec4574b59c0b1ca2d3a4e7319 is an example

xprazak2:
I just found a bug, please do not merge yet...

xprazak2:
I forgot to actually load the settings in test env before using them in tests... fixed now.

test/fixtures/templates_settings.yml
```diff
@@ -0,0 +1,40 @@
+templates_verbose:
```
GregSutcliffe: I'm not a huge fan of fixtures - it was me that did the initial work to move to FactoryGirl and we're *still* trying to get rid of the rest :P. So, I don't really want to add *more* fixtures if we can avoid it.

Is there any reason not to use https://github.com/theforeman/foreman/blob/develop/test/factories/settings.rb?

app/models/setting/templates.rb
```diff
@@ -0,0 +1,29 @@
...
+      ].compact.each { |s| self.create! s.update(:category => "Setting::Templates") }
+    end
+
+    true
+  end
+
+  def validate_templates_associate(record)
```
GregSutcliffe: Can we use something like https://github.com/theforeman/foreman/blob/develop/app/models/setting/provisioning.rb#L26 to create a selector rather than a textbox+validation?

test/support/foreman_test_helper_additions.rb
```diff
@@ -0,0 +1,21 @@
+require "#{ForemanTemplates::Engine.root}/test/support/fixture_support"
```
GregSutcliffe: These two files can probably go away if you use factories too - since the factory is in core, you can just use it directly in tests, eg https://github.com/theforeman/foreman_default_hostgroup/blob/develop/test/unit/default_hostgroup_test.rb#L15-L17

xprazak2:
I replaced fixtures with factories, there was no special reason why I was using them. There is also a selector for association setting instead of input field now.

GregSutcliffe:
Looks pretty good, I'll try it out later today.

One thing that might come up - this will create a Settings tab called "Templates" which might be a little misleading (since people might think it refers to the `config_templates` controller / models rather than this plugin.

Do we have anything in core that allows us to rename the tab or add some description?

xprazak2:
It does not seem like we do, but I will look more deeply.

GregSutcliffe:
Yeah, I wasn't expecting so. Don't spend too long on it :)

stbenjam:
It's generated from the class name, so you could call it something other than Settings::Templates, like Settings::TemplateSync or whatever

xprazak2:
I have a solution in progress... https://github.com/theforeman/foreman/pull/3976

GregSutcliffe:
@stbenjam ah, thats a good point, probably the easiest solution. @xprazak2 not too keen on tying this to a commit in foreman core if we don't have to (we can't release this until 1.14 then)

xprazak2:
OK, I will rename setting category to TemplateSync then.

GregSutcliffe:
@xprazak2 thanks, I'll be happy to do a new release with `humanized_category` once but 1.14 is out, but for now I think this is a good compromise.

xprazak2:
No problem, setting class is now `Setting::TemplateSync`.

GregSutcliffe:
Quick test here - some of the values seem incorrectly initialised (on 1.13-stable, if it matters):

spectacle hz5501

The collections there, as you can see when you edit:

spectacle vc5501

It seems correct in the console:

```
irb(main):008:0> Setting[:template_sync_associate]
=> "New"
```

so just some UI weirdness?

GregSutcliffe:
(misclick on "Close & comment", sorry for the noise)

sean797:
Also I think we should validate `template_sync_associate` setting in case its set via the API?

xprazak2:
UI weirdness was caused by my typo, `template_sync_associate` should display correctly now, I also added a validation. I do not think we need to validate `template_sync_verbose` and `template_sync_negate` because booleans are [validated by default](https://github.com/theforeman/foreman/blob/develop/app/models/setting.rb#L34).

sean797:
Nice @xprazak2! One small thing, now that sync settings is in the Settings page can we make sure its fairly obvious people can still use the rake task and specify the options there, for new users it may not be obvious they can sync from multiple git repos by specifying the options on the command line. I'm not sure the best way to do this though..

xprazak2:
I changed readme a bit so that new users are aware that passing options to a Rake task overrides the defaults from settings.

xprazak2:
@GregSutcliffe could you take another look pls?

GregSutcliffe:
@xprazak2 yeah, it's on my to-do, might not get to it until later this week though.

app/models/setting/template_sync.rb
```diff
@@ -0,0 +1,37 @@
...
+      ].compact.each { |s| self.create! s.update(:category => "Setting::TemplateSync") }
+    end
+
+    true
+  end
+
+  def validate_template_sync_associate(record)
+    values = record.class.associate_types.keys
+    if record.value && !values.include?(record.value)
+      record.errors[:base] << (_("template_sync_associate must be one of %s") % values)
```
ares: you probably need to do `values.join(', ')`, I'm getting just "always" here

ares:
Works great, just one small comment regarding validation message, the I think it's ready for merge.

xprazak2:
I made a change to fix the validation message.

ares:
:+1: @GregSutcliffe do you want to do another round of review or should I merge?

GregSutcliffe:
I'm happy if you're happy @ares - code looks fine. If I get chance I'll test it manually, but I don't think it's a blocker ;)

Do you want me to do a release, or is there more to come that's worth waiting for?

ares:
Thanks @xprazak2, merging. @GregSutcliffe I believe there's more to come soon from @xprazak2 so I'd wait with a release few more days.

GregSutcliffe:
Thanks @xprazak2!

PR #35, opened 2016-11-21
Fixes #16984 - Import templates from filesystem

xprazak2:


app/services/foreman_templates/template_importer.rb
```diff
@@ -25,6 +25,21 @@ def initialize(args = {})
     end
 
     def import!
+      if @repo.start_with?('http://', 'https://', 'git://')
+        import_from_git
+      elsif repo.start_with?('/')
```
GregSutcliffe: Should we allow a relative dir here? Perhaps check if it starts with "." ?

GregSutcliffe:
Seems simple enough - no tests though. Can you add some please?

app/services/foreman_templates/template_importer.rb
ares: is it a good idea? I'd like to trigger sync from Foreman UI later, what would pwd be in case of foreman user? What if the task is run from cron? I think absolute paths is not a big limitation and can save us from few headaches :-)

ares:
I asked @xprazak2 to push this without tests so I can see how he plans to support pure directory imports so I can write my export PR in a compatible way. I think tests are on roadmap.

app/services/foreman_templates/template_importer.rb
GregSutcliffe: I'm not saying *default* to relative - absolute should definitely be the recommended use. However, I can think of a few situations where the absolute path may not be known - automation in general (Jenkins, tests, other CI pipelines) tend to use temporary directories. Also, if you're doing any kind of binding (chroots, or cross-mounted filesystems) then again it can be tricky to figure out the absolute path.

I'm open to leaving it absolute-only for now, but I can see people wanting it later. Does it hurt to support relative paths, along with a "here be dragons" warning?

GregSutcliffe:
@ares makes sense. No pressure, just registering that tests will be needed :)

app/services/foreman_templates/template_importer.rb
ares: Hm. The only concern is number of irc questions :-) thinking about this more, also paths like ~foreman/ or ~jenkins/ would be useful. Maybe we can keep filesystem as `else` and try to validate paths later if it seems necessary. Documentation is the key though. After all these changes, we need to write the manual.

GregSutcliffe: Yeah, `~` is useful, good point. Falling back to file-based sounds reasonable - the error message can become `Using file-based import, but couldn't find #{File.expand_path(@repo)}` (or something like that) to show where the code is searching. 

+1 for a manual, but you might want at least a short note in the README for now, until that gets written. Don't spend too long on it ;)

xprazak2:
I added tests, moved `import_from_files` into else branch, and checking if the given path actually exists. Relative paths are handled with `File.expand_path`

test/unit/template_importer_test.rb
```diff
@@ -225,18 +225,57 @@ def self.import!(name, text, _metadata)
       end
     end
 
+    test 'should import files from git' do
+      setup_settings
+      imp = ForemanTemplates::TemplateImporter.new
```
GregSutcliffe: Oh I see, you're not using `@importer` - but that means you're cloning from the real community-templates repo instead of from the fixtures....

ares: @xprazak2 could you look at it? I'd love to see this getting in soon

xprazak2: It should be ok now, all I want to test here is that the right type of import source is chosen when repo is specified with https://. `import_from_git` is stubbed in test, so there is no calling out to external repos.

GregSutcliffe: Ah cool, that'll simplify tests in the future too, since it can use `import_from_files`. I'll give it a quick test but it's looking good!

GregSutcliffe:
A quick test locally seems to work - but `import_from_git` is untested now. However, `parse_files` is tested via `import_from_files` so I'm inclined to ignore it.

This takes coverage down from 73.56 % to 68.75% - we should probably add a TODO to improve coverage in `app/services/foreman_templates/template_importer.rb`

@ares any concerns there? I think I'm happy with it.

ares:
I think we'll improve test coverage with upcoming changes. :+1: from me too, so merging. Thanks @xprazak2.

PR #36, opened 2016-11-21
Fixes #17417 - adds exporting task

ares:
Requires https://github.com/theforeman/foreman/pull/4031 but other than that, ready for review. I didn't cover it in readme since I'd like to write a proper manual for the plugin and start linking there. If you want to test, run `rake templates:export`. Just be careful, currently it clones remote repo, does changes, creates commit and pushes to the origin. After discussion with @xprazak2 we decided to leave it like this until there's a filesystem support (https://github.com/theforeman/foreman_templates/pull/35). Later we'd like to introduce repos as first class citizen object that users could configure separately.

### DIFF
--- a/app/models/setting/template_sync.rb
+++ b/app/models/setting/template_sync.rb
@@ -25,7 +25,7 @@ class Setting::TemplateSync < Setting
       [
         self.set('template_sync_verbose', N_('Choose verbosity for Rake task importing templates'), false, N_('Verbosity')),
         self.set('template_sync_associate', N_('Associate templates to OS'), 'new', N_('Associate'), nil, { :collection => Proc.new { self.associate_types } }),
-        self.set('template_sync_prefix', N_('The string all imported or exported templates should begin with'), "Community ", N_('Prefix')),
+        self.set('template_sync_prefix', N_('The string all imported templates should begin with'), "Community ", N_('Prefix')),
         self.set('template_sync_dirname', N_('The directory within the Git repo containing the templates'), '/', N_('Dirname')),
         self.set('template_sync_filter', N_('Import or export names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
         self.set('template_sync_repo', N_('Default Git repo to sync from'), 'https://github.com/theforeman/community-templates.git', N_('Repo')),
--- a/app/models/setting/template_sync.rb
+++ b/app/models/setting/template_sync.rb
@@ -7,6 +7,15 @@ class Setting::TemplateSync < Setting
     }
   end
 
+  def self.metadata_export_mode_types
+    {
+      'refresh' => _('Refresh'),
+      'keep' => _('Keep'),
+      'remove' => _('Remove')
+    }
+  end
+
+
   def self.load_defaults
     return unless super
 
@@ -16,12 +25,13 @@ class Setting::TemplateSync < Setting
       [
         self.set('template_sync_verbose', N_('Choose verbosity for Rake task importing templates'), false, N_('Verbosity')),
         self.set('template_sync_associate', N_('Associate templates to OS'), 'new', N_('Associate'), nil, { :collection => Proc.new { self.associate_types } }),
-        self.set('template_sync_prefix', N_('The string all imported templates should begin with'), "Community ", N_('Prefix')),
+        self.set('template_sync_prefix', N_('The string all imported or exported templates should begin with'), "Community ", N_('Prefix')),
         self.set('template_sync_dirname', N_('The directory within the Git repo containing the templates'), '/', N_('Dirname')),
-        self.set('template_sync_filter', N_('Import names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
+        self.set('template_sync_filter', N_('Import or export names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
         self.set('template_sync_repo', N_('Default Git repo to sync from'), 'https://github.com/theforeman/community-templates.git', N_('Repo')),
-        self.set('template_sync_negate', N_('Negate the search'), false, N_('Negate')),
-        self.set('template_sync_branch', N_('Default branch in Git repo'), nil, N_('Branch'))
+        self.set('template_sync_negate', N_('Negate the prefix (for purging) / filter (for importing/exporting)'), false, N_('Negate')),
+        self.set('template_sync_branch', N_('Default branch in Git repo'), nil, N_('Branch')),
+        self.set('template_sync_metadata_export_mode', N_('Default metadata export mode, refresh re-renders metadata, keep will keep existing metadata, remove exports template withou metadata'), 'refresh', N_('Metadata export mode'), nil, { :collection => Proc.new { self.metadata_export_mode_types } })
       ].compact.each { |s| self.create! s.update(:category => "Setting::TemplateSync") }
     end
 
@@ -32,6 +42,13 @@ class Setting::TemplateSync < Setting
     values = record.class.associate_types.keys
     if record.value && !values.include?(record.value)
       record.errors[:base] << (_("template_sync_associate must be one of %s") % values.join(', '))
+    end
+  end
+
+  def validate_template_sync_metadata_export_mode(record)
+    values = record.class.metadata_export_mode_types.keys
+    if record.value && !values.include?(record.value)
+      record.errors[:base] << (_("template_sync_metadata_export_mode must be one of %s") % values.join(', '))
     end
   end
 end
--- a/app/services/foreman_templates/template_exporter.rb
+++ b/app/services/foreman_templates/template_exporter.rb
@@ -26,15 +26,27 @@ module ForemanTemplates
       git_repo = Git.clone(@repo, @dir)
       logger.debug "cloned #{@repo} to #{@dir}"
       branch = @branch ? @branch : get_default_branch(git_repo)
-      git_repo.checkout(branch) if branch
+      # either checkout to existing or create a new one and checkout afterwards
+      if branch
+        if git_repo.is_branch?(branch)
+          git_repo.checkout(branch)
+        else
+          git_repo.branch(branch).checkout
+        end
+      end
 
       dump_files!
-
       git_repo.add
-      logger.debug "commiting changes in cloned repo"
-      git_repo.commit "Templates export made by Foreman user #{User.current.try(:login) || User::ANONYMOUS_ADMIN}"
-      logger.debug "pushing to branch #{branch} at origin #{@repo}"
-      git_repo.push 'origin', branch
+
+      if git_repo.status.added.any?
+        logger.debug "committing changes in cloned repo"
+        git_repo.commit "Templates export made by Foreman user #{User.current.try(:login) || User::ANONYMOUS_ADMIN}"
+
+        logger.debug "pushing to branch #{branch} at origin #{@repo}"
+        git_repo.push 'origin', branch
+      else
+        logger.debug 'no change detected, skipping the commit and push'
+      end
     ensure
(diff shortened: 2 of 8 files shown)


app/models/setting/template_sync.rb
```diff
@@ -7,21 +7,31 @@ def self.associate_types
...
+      'remove' => _('Remove')
+    }
+  end
+
+
   def self.load_defaults
     return unless super
 
-    %w(template_sync_filter template_sync_branch).each { |s| Setting::BLANK_ATTRS << s }
+    %w(template_sync_filter template_sync_branch ).each { |s| Setting::BLANK_ATTRS << s }
```
GregSutcliffe: Nitpick: is this extra whitespace intentional?

app/models/setting/template_sync.rb
```diff
@@ -7,21 +7,31 @@ def self.associate_types
...
-        self.set('template_sync_prefix', N_('The string all imported templates should begin with'), "Community ", N_('Prefix')),
+        self.set('template_sync_prefix', N_('The string all imported or exported templates should begin with'), "Community ", N_('Prefix')),
         self.set('template_sync_dirname', N_('The directory within the Git repo containing the templates'), '/', N_('Dirname')),
-        self.set('template_sync_filter', N_('Import names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
+        self.set('template_sync_filter', N_('Import or export names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
         self.set('template_sync_repo', N_('Default Git repo to sync from'), 'https://github.com/theforeman/community-templates.git', N_('Repo')),
-        self.set('template_sync_negate', N_('Negate the search'), false, N_('Negate')),
-        self.set('template_sync_branch', N_('Default branch in Git repo'), nil, N_('Branch'))
+        self.set('template_sync_negate', N_('Negate the prefix search for purging, filter for importing or exporting'), false, N_('Negate')),
```
GregSutcliffe: Nitpick: This can probably be worded better - perhaps "Negate the prefix (for purging) / filter (for importing/exporting)" ?

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -24,6 +27,23 @@ namespace :templates do
     puts results.join("\n")
   end
 
+  task :import => :sync
```
GregSutcliffe: Nitpick: Shouldn't this the other way around? :sync is deprecated, so the code should now be at :import

app/services/foreman_templates/template_exporter.rb
```diff
@@ -0,0 +1,78 @@
...
+        filename = File.join(current_dir, get_template_name(template))
+        File.open(filename, 'w+') do |file|
+          logger.debug "Writing to file #{filename}"
+          bytes = file.write template.public_send(export_method)
+          logger.debug "finished writing #{bytes}"
+        end
+      end
+    end
+
+    def get_template_name(template)
```
GregSutcliffe: `name` is a bit overloaded, using `get_template_filename` would be clearer

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -24,6 +27,23 @@ namespace :templates do
...
+      repo: ENV['repo'],
+      branch: ENV['branch'],
+      prefix: ENV['prefix'],
+      dirname: ENV['dirname'],
+      filter: ENV['filter'],
+      # associate: ENV['associate'],
+      metadata_export_mode: ENV['metadata_export_mode'],
+    }).export!
+
+    puts 'Export finished'
```
GregSutcliffe: I see the internal code logs using `logger` - but just this single line for rake. Can we get output parity here (at least using `@verbose`)?

GregSutcliffe: Ignore me, seems to work just fine now that I try it ;)

GregSutcliffe:
@ares: Should this work with local git dirs? I tried this:

```
mkdir /tmp/foo && cd /tmp/foo && git init && cd - 
bundle exec rake templates:export verbose=true repo="/tmp/foo"
```
and it fails with
```
...
2016-11-23T15:55:35  [app] [D] Writing to file /tmp/d20161123-9598-2oioni/provisioning_templates/PXELinux/Community greg_alterator_default_pxelinux.erb
2016-11-23T15:55:35  [app] [D] finished writing 640
2016-11-23T15:55:35  [app] [D] commiting changes in cloned repo
2016-11-23T15:55:35  [app] [D] pushing to branch  at origin /tmp/foo
rake aborted!
Git::GitExecuteError: git '--git-dir=/tmp/d20161123-9598-2oioni/.git' '--work-tree=/tmp/d20161123-9598-2oioni' push 'origin' ''  2>&1:fatal: remote part of refspec is not a valid name in 
/home/greg/github/foreman_templates/app/services/foreman_templates/template_exporter.rb:21:in `export!'
/home/greg/github/foreman_templates/lib/tasks/foreman_templates_tasks.rake:42:in `block (2 levels) in <top (required)>'
/home/greg/.rbenv/versions/2.3.1/bin/bundle:23:in `load'
/home/greg/.rbenv/versions/2.3.1/bin/bundle:23:in `<main>'
Tasks: TOP => templates:export
(See full trace by running task with --trace)
```

ares:
It should, you need to have at least one commit there though. Also note that the branch must not be checked out before pushing (e.g. checkout to different branch than you push to).

GregSutcliffe:
No, it still bombs. I'm doing this:

```
rm -rf /tmp/foo ; mkdir /tmp/foo ; cd /tmp/foo
git init ; touch foo ; and git add foo ; and git ci -m "Init" ; and git co -t master -b devel ; and cd -
Initialized empty Git repository in /tmp/foo/.git/
bundle exec rake templates:export verbose=true repo="/tmp/foo" prefix="Greg "
```

Same error, I think:
```
2016-11-23T16:05:08  [app] [D] commiting changes in cloned repo
2016-11-23T16:05:08  [app] [D] pushing to branch  at origin /tmp/foo
rake aborted!
Git::GitExecuteError: git '--git-dir=/tmp/d20161123-10907-176cv2s/.git' '--work-tree=/tmp/d20161123-10907-176cv2s' push 'origin' ''  2>&1:fatal: remote part of refspec is not a valid name in 
/home/greg/github/foreman_templates/app/services/foreman_templates/template_exporter.rb:21:in `export!'
/home/greg/github/foreman_templates/lib/tasks/foreman_templates_tasks.rake:42:in `block (2 levels) in <top (required)>'
```

GregSutcliffe:
The prefixing is a bit weird too. I used `templates:import prefix='Greg '` to import my templates, then when I export with the same prefix, I get:

```
/tmp/d20161123-9799-1y90ler/provisioning_templates/PXELinux/Greg greg_alterator_default_pxelinux.erb
```

See how it's kept the import prefix (but with lowercase and whitespace->underscore) but also added the prefix again. This is also not handled in the `name:` field of the dumped text (it's `Greg Alterator ...`)

The filename is mostly cosmetic, but (assuming you automate the import step with a known prefix) the `name` metadata would lead to an increasing number of prefixes on each import/export cycle, since "Greg Alterator..." would get prefixed with "Greg " again on the second import.

Both import and export should probably check if the string already starts with the prefix and drop it if so.

ares:
@GregSutcliffe the error you're receiving is caused by the fact you haven't created a development branch in your repo. We default to it since we assume the repo is community-templates. You can also override the branch in settings but it must exist in target repo. We should probably add some layer for better error detection and reporting, but that's out of scope for this PR I'm afraid as this is the same for importing.

Regarding the prefix, I think we just need to stop importing double prefixes. During the export, prefix is already part of the name and you're right, it does not make sense to add it to the filename since normally it should already be there.

I fixed all your comments, please take another look. I'd also like to add pure FS export support in this PR so please do not merge yet. I hope to find some time today to finish it.

ares:
Now with FS support, so cancelling the "do not merge" status :-)

GregSutcliffe:
@ares we're closer but not done yet :)

* The git thing makes sense - I've tested with a real repo and it seems good. For reference I used `bundle exec rake templates:export verbose=true repo="https://github.com/GregSutcliffe/community-templates.git" branch="disk" prefix="Greg "` and it resulted in https://github.com/GregSutcliffe/community-templates/tree/disk
* File based export is working fine too 🎉 
* Prefixes are still broken

So you've only added `auto_prefix` to `import` as far as I can tell. But you can still specify it on export (as in my example above). So if you start with a template called "foo_bar" it exports as "Greg foo_bar" and then imports as "Greg foo_bar". After one round it seems to stabilise - I don't see "Greg Greg greg_foo_bar" or anything like that - but it's still a bit weird.

Does it make sense to drop the `prefix` option from the export task? We're just trying to represent the state of the Foreman DB on disk, so perhaps we shouldn't prefix the names.

There's also two other small things I'll comment on inline.

app/services/foreman_templates/template_importer.rb
```diff
@@ -69,8 +65,12 @@ def parse_files!
...
-        name     = [@prefix, name].compact.join
-        next if @filter && !name.match(/#{@filter}/i)
+        name     = auto_prefix(name)
+        if @filter
+          matching = name.match(/#{@filter}/i)
+          matching = !matching if @negate
+          next if !matching
+        end
 
         raise MissingKindError unless metadata['kind']
```
GregSutcliffe: The export task isn't writing `kind` for ptables or snippets, so the subsequent import fails. We should probably remove this check now that we're moving to `model` - but we might want to add `model` to the export task so that we always hit line 79 below

app/services/foreman_templates/template_exporter.rb
```diff
@@ -0,0 +1,92 @@
...
+      dump_files!
+    end
+
+    def export_to_git
+      @dir = Dir.mktmpdir
+
+      git_repo = Git.clone(@repo, @dir)
+      logger.debug "cloned #{@repo} to #{@dir}"
+      branch = @branch ? @branch : get_default_branch(git_repo)
+      git_repo.checkout(branch) if branch
```
GregSutcliffe: Is there any chance to create the branch if we specify a non-existant one? Currently it fails unless you give a valid branch, and it felt a bit weird to have to clone my repo separately just to add a target for exporting. This is not a blocker and can be done in another PR, just curious

ares: easy enough to do it in this PR

ares:
I kept changes in separate commit for easier reviewing. I think it should address all recent comments. Please take another look.

ares:
https://github.com/theforeman/foreman/pull/4098 has been merged so exported templates will have model in the yaml if it runs with 1.15

ares:
There's still a chance to get this merged in 2016 :-)

GregSutcliffe:
Sorry for the delay @ares 

Testing pretty well here, the only minor issue/question left is what to do with prefixed imports. Currently you get something like this:

```
> select name, template from templates;
...
Greg Preseed default|<%#
name: Preseed default
snippet: false
model: Ptable
os_family: Debian
%>
```

Note how the Prefix goes in the name but not the metadata. Now, when this is *exported* it'll be fixed (assuming metadata==refresh) but it leaves it slightly inconsistent in the meantime.

I don't think it's a big issue, so let's [test] one more time, and then merge, and we can deal with edge cases as we go.

ares:
I didn't find a better way. I agree, lets hear from users how they would prefer to solve it. If this is commonly used thing we might have prefix as first class citizen attribute in templates.

GregSutcliffe:
@ares looks like you've got a minor test failure. +1 when it's green ;)

ares:
thou shalt not fail

ares:
@GregSutcliffe :green_apple:

GregSutcliffe:
Squashed and merged, ty @ares for the hard work and patience :)

PR #37, opened 2016-12-05
Fixes #17583 - Provide an API to initiate a sync

sean797:
You should be able to test it with:
```
curl -H "Accept:application/json,version=2" -H "Content-Type:application/json" -u admin:changeme -k https://centos7-devel.example.com/api/v2/templates/import -X POST -d ''
```

### DIFF
--- a/app/controllers/api/v2/template_controller.rb
+++ b/app/controllers/api/v2/template_controller.rb
@@ -1,0 +1,28 @@
+module Api
+  module V2
+    class TemplateController < ::Api::V2::BaseController
+
+      api :POST, "/template/import/", N_("Initiate Import")
+      param :repo, String, :required => false, :desc => N_("Import templates from a different repo.")
+      param :branch, String, :required => false, :desc => N_("Branch in Git repo.")
+      param :prefix, String, :required => false, :desc => N_("The string all imported templates should begin with.")
+      param :dirname, String, :required => false, :desc => N_("The directory within the git tree containing the templates.")
+      param :filter, String, :required => false, :desc => N_("Import names matching this regex (case-insensitive; snippets are not filtered).")
+      param :negate, :bool, :required => false, :desc => N_("Negate the prefix (for purging).")
+      param :associate, String, :required => false, :desc => N_("Associate to OS's, Locations & Organizations. Options are: always, new or never.")
+      def import
+        results = ForemanTemplates::TemplateImporter.new({
+          verbose:   params['verbose'],
+          repo:      params['repo'],
+          branch:    params['branch'],
+          prefix:    params['prefix'],
+          dirname:   params['dirname'],
+          filter:    params['filter'],
+          associate: params['associate'],
+          negate:    params['negate'],
+        }).import!
+        render :json => {:message => results}
+      end
+    end
+  end
+end
--- a/lib/foreman_templates/engine.rb
+++ b/lib/foreman_templates/engine.rb
@@ -15,7 +15,16 @@ module ForemanTemplates
 
     initializer 'foreman_templates.register_plugin', :before => :finisher_hook do
       Foreman::Plugin.register :foreman_templates do
-        requires_foreman '>= 1.14'
+        requires_foreman '>= 1.15'
+
+        apipie_documented_controllers ["#{ForemanTemplates::Engine.root}/app/controllers/api/v2/*.rb"]
+
+        security_block :templates do
+          permission :import_templates, {
+            :"api/v2/template" => [:import]
+          }, :resource_type => 'Template'
+        end
+        add_all_permissions_to_default_roles
       end
     end
 
--- a/test/functional/api/v2/template_controller_test.rb
+++ b/test/functional/api/v2/template_controller_test.rb
@@ -1,0 +1,10 @@
+require 'test_plugin_helper'
+
+class Api::V2::TemplateControllerTest < ActionController::TestCase
+
+  test "should import from git" do
+    post :import, { 'repo': "#{ForemanTemplates::Engine.root}/test/templates/core", 'prefix': ''}
+    assert_response :success
+  end
+
+end
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *.swp
 .idea
+Gemfile.lock
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,0 +1,13 @@
+Rails.application.routes.draw do
+
+  namespace :api, :defaults => {:format => 'json'} do
+    scope "(:apiv)", :module => :v2, :defaults => {:apiv => 'v2'}, :apiv => /v1|v2/, :constraints => ApiConstraints.new(:version => 2) do
+      resources :templates, :controller => :template, :only => :import do
+        collection do
+          post 'import'
+        end
+      end
+    end
+  end
+
+end


GregSutcliffe:
See, this is why I moved the code to a service class ;)

Looks ok to me, thanks @sean797 - @ares?

app/controllers/api/v2/templates_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    api :GET, "/templates/sync/", N_("Initiate Sync")
+    param :repo, String, :required => false, :desc => N_("Sync templates from a different repo.")
+    param :branch, String, :required => false, :desc => N_("Branch in Git repo.")
+    param :prefix, String, :required => false, :desc => N_("The string all imported templates should begin with.")
+    param :dirname, String, :required => false, :desc => N_("The directory within the git tree containing the templates.")
+    param :filter, String, :required => false, :desc => N_("Import names matching this regex (case-insensitive; snippets are not filtered).")
+    param :associate, String, :required => false, :desc => N_("Associate to OS's, Locations & Organizations. Options are: always, new or never.")
+    def sync
+      results = ForemanTemplates::TemplateImporter.new({
+        verbose:   params['verbose'],
```
ares: please avoid the white-space aligning, it's hard to review changes when new longer key is added

lib/foreman_templates/engine.rb
```diff
@@ -16,6 +16,14 @@ class Engine < ::Rails::Engine
     initializer 'foreman_templates.register_plugin', :before => :finisher_hook do
       Foreman::Plugin.register :foreman_templates do
         requires_foreman '>= 1.12'
+
+        security_block :templates do
+          permission :sync_templates, {
+            :templates => [:sync],
+            :"api/v2/templates" => [:sync]
+          }, :resource_type => 'templates' 
+        end
```
ares: it would be great to add this permission to Manager role

ares:
In general I like it, before we merge 'sync_templates' though I'd like to finish exporting task first since it's going to rename the action to import to avoid. That also applies to action name. I left few comments in the code too.

GregSutcliffe:
@ares That's true, I'm good with the action rename - do you have an ETA on that? I don't want to hold up a useful PR for too long, especially if the API is useful downstream.

ares:
@GregSutcliffe I think it now waits for your review and merge :-) I'm happy to help with this PR after that.

sean797:
Made a couple of changes @ares. 
- Renamed sync to import.
- Permission is now added to the Manager role.

sean797:
[test]

lib/foreman_templates/engine.rb
```diff
@@ -16,6 +16,17 @@ class Engine < ::Rails::Engine
...
+
+        security_block :templates do
+          permission :import_templates, {
+            :template => [:import],
+            :"api/v2/template" => [:import]
+          }, :resource_type => 'Template' 
+        end
+
+        role = Role.where(:name => 'Manager').first
+        role.add_permissions!('import_templates') unless role.permissions.find_by_name('import_templates')
```
sean797: @ares Hows the best way to do this? Looks like its fails because the import_template permissions isn't registered yet. I wanted to avoid a db migration but it looks like the only way to me..

ares: There's a DSL for defining permissions, see [this](https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L53) for example. Roles are defined like [this](https://github.com/theforeman/foreman_openscap/blob/master/lib/foreman_openscap/engine.rb#L109) but @xprazak2 works on better API for plugins in https://github.com/theforeman/foreman/pull/4168 (maybe your input as plugin developer would move the discussion on foreman-dev mailing list further)

ares: sorry I missed you already use the `permission`  few lines above :-) the `role` should work though I think

sean797: Im just trying to add the permission to the manager role really. it does actually work on my devel box but test seem to fail :(

ares: So what happens is that this code get executed whenever the rails environment is loaded which happens also during migration. Since the way we run tests we do the plugin migrations before the DB of foreman core is seeded, the role does not exist yet so it fails on `role` variable being nil.

You can wrap the code in if condition to skip adding permission if the role does not exist. Note that the same code will be executed later when db is seeded or rails server starts so the permission will be added later when DB is ready.

```ruby
role = Role.where(:name => 'Manager').first
if role
  role.add_permissions!('import_templates') unless role.permissions.find_by_name('import_templates')
end
```

adding a explaining comment there would be a good idea. Once https://github.com/theforeman/foreman/pull/4168 is resolved we could update this code accordingly.

ares:
Hm, I see it didn't help, I'll try to figure out better way, we can probably do the check like we do with Settings but maybe plugin DSL should provide that. I'll play with it and see how it could work.

ares:
@sean797 I sent a PR against you fork. If you merge it, my comment will appear here as well. That should fix the tests. Once we have better solution in core, we can refactor this hack.

sean797:
@ares Okay, yea I guess that will work 👍 Thanks!

ares:
Hm these failures seems weird and unrelated but I didn't have a chance to look closer.

sean797:
@ares I think test are failing due to something theforeman/foreman@567cb9f1ef5098d8337403be23784c2c97e491f1 any chance you can take a look? I not sure what it is

sean797:
[test]

https://github.com/theforeman/foreman/commit/807d975add1d518350f3e8b45f76bc2dfdc5d7ba should have fixed the failure.

sean797:
Yippy! Should be good to merge, tested passed 👍 

Thanks @ares & @GregSutcliffe !

app/controllers/api/v2/template_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+  module V2
+    class TemplateController < ::Api::V2::BaseController
+
+      api :GET, "/template/import/", N_("Initiate Import")
+      param :repo, String, :required => false, :desc => N_("Import templates from a different repo.")
+      param :branch, String, :required => false, :desc => N_("Branch in Git repo.")
+      param :prefix, String, :required => false, :desc => N_("The string all imported templates should begin with.")
+      param :dirname, String, :required => false, :desc => N_("The directory within the git tree containing the templates.")
+      param :filter, String, :required => false, :desc => N_("Import names matching this regex (case-insensitive; snippets are not filtered).")
+      param :associate, String, :required => false, :desc => N_("Associate to OS's, Locations & Organizations. Options are: always, new or never.")
```
ares: `negate` option is missing, would you mind adding it?

app/controllers/api/v2/template_controller.rb
```diff
@@ -0,0 +1,26 @@
+module Api
+  module V2
+    class TemplateController < ::Api::V2::BaseController
+
+      api :GET, "/template/import/", N_("Initiate Import")
```
ares: it's listed here as GET but in routes, it's `post` which makes more sense to me

lib/foreman_templates/engine.rb
```diff
@@ -15,7 +15,17 @@ class Engine < ::Rails::Engine
...
     initializer 'foreman_templates.register_plugin', :before => :finisher_hook do
       Foreman::Plugin.register :foreman_templates do
-        requires_foreman '>= 1.14'
+        requires_foreman '>= 1.15'
+
+        apipie_documented_controllers ["#{ForemanTemplates::Engine.root}/app/controllers/api/v2/*.rb"]
+
+        security_block :templates do
+          permission :import_templates, {
+            :template => [:import],
```
ares: where is this controller/action defined? I only see API being added

test/functional/api/v2/template_controller_test.rb
```diff
@@ -0,0 +1,12 @@
+require 'test_plugin_helper'
+
+class Api::V2::TemplateControllerTest < ActionController::TestCase
+
+  test "should import" do
+    post :import, { 'repo':'/does-not-exist' }
+    response = ActiveSupport::JSON.decode(@response.body)
+    assert_equal response['error']['message'], "Using file-based synchronization, but couldn't find /does-not-exist"
+    assert_response :error
+  end
```
ares: a test for positive path would be appreciated

ares:
Thanks @sean797, I left few hopefully small comments. If you get to it, I'd like to have this in upcoming release.

test/functional/api/v2/template_controller_test.rb
sean797: I done it this way because if we send it a git path its takes a fair amount of time to clone & import also there's potential for unrelated failures? 

But I'm happy to add a git URL as well..

ares:
Seems like there's some problem with the test, maybe the path must be constructed like `ForemanTemplates::Engine.root + 'test/...'`? http://ci.theforeman.org/job/test_plugin_pull_request/3610/database=mysql,label=fast,ruby=2.2/testReport/junit/(root)/Api__V2__TemplateControllerTest/test_0001_should_import_from_git/

test/functional/api/v2/template_controller_test.rb
ares: plain directory is fine, if you still have also "negative" path, feel free to keep it there too, both are testing different code paths

test/functional/api/v2/template_controller_test.rb
```diff
@@ -0,0 +1,10 @@
+require 'test_plugin_helper'
+
+class Api::V2::TemplateControllerTest < ActionController::TestCase
+
+  test "should import from git" do
+    post :import, { 'repo': File.expand_path(File.join('..', '..', '..', '..', '..'), __FILE__) + '/test/templates/core'}
```
ares: Why not using `ForemanTemplates::Engine.root + '/test/templates/core'`, I think that should be equivalent but less prone to errors.

sean797: because `ForemanTemplates::Engine.root`was `nil` when I quickly tested it. Its working now, maybe i had a typo before..

sean797:
@ares tests passed, I had send `prefix: '' ` in the test as without it `@prefix` is `nil` in app/services/foreman_templates/template_importer.rb in the test environment.

ares:
> tests passed, I had send prefix: '' in the test as without it @prefix is nil in app/services/foreman_templates/template_importer.rb in the test environment.

makes sense, I think that's because the setting is not seeded in test environment, I'll give it a try and if it works without specifying the prefix I'll merge

app/controllers/api/v2/template_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+      param :negate, :bool, :required => false, :desc => N_("Negate the prefix (for purging).")
+      param :associate, String, :required => false, :desc => N_("Associate to OS's, Locations & Organizations. Options are: always, new or never.")
+      def import
+        results = ForemanTemplates::TemplateImporter.new({
+          verbose:   params['verbose'],
+          repo:      params['repo'],
+          branch:    params['branch'],
+          prefix:    params['prefix'],
+          dirname:   params['dirname'],
+          filter:    params['filter'],
```
ares: I see you added `negate` to API, I think it also needs to be used here

sean797: Sorry that was super silly of me. Updated!

ares:
Thanks @sean797, I'd like to do one more release before we bump Foreman dependency. But don't worry I'll do next release with this one just after it, the plugin needs some fixes for 1.15 anyway.

ares:
Thanks @sean797, merging.

ares:
@sean797 I'd appreciate if you documented the API use (e.g. example of curl, the required permission) at https://github.com/theforeman/theforeman.org/tree/gh-pages/plugins/foreman_templates once we copy manual for 0.5 release

sean797:
@ares sure! Ping me when we have the 0.5 copy.

ares:
@sean797 the manual was cloned for 5.0, could you please update the chapter 5. API accordingly? the file you need to update is here https://github.com/theforeman/theforeman.org/blob/gh-pages/plugins/foreman_templates/5.0/index.md

PR #44, opened 2017-04-12
Fixes #19260 - add a clean up task

ares:


app/services/foreman_templates/uninstaller.rb
```diff
@@ -0,0 +1,13 @@
+module ForemanTempaltes
```
GregSutcliffe: typo in module name

lib/tasks/foreman_templates_tasks.rake
```diff
@@ -57,6 +57,13 @@ namespace :templates do
     }).purge!
   end
 
+  desc 'Clean data created by this plugin, this will permanently delete the data!'
```
GregSutcliffe: Overly dramatic? It implies it'll delete the *templates* which clearly isn't true. Perhaps:
```
Clean up for uninstallation- this removes all custom Settings added by the plugin
```
or similar?

app/services/foreman_templates/uninstaller.rb
```diff
@@ -0,0 +1,13 @@
...
+  class Cleaner
```
GregSutcliffe: Cleaner != "uninstaller.rb" so it isn't autoloaded. Renaming the file to "cleaner.rb" works

GregSutcliffe:
Tested with my review changes, works fine. ACK once updated.

lib/tasks/foreman_templates_tasks.rake
ares: Hm, could you find some more generic message? I'd like this to be generic for future. Currently it only deletes Settings but the concept should be also for any default data coming from this plugin.

app/services/foreman_templates/uninstaller.rb
ares: good catch, I did the last second rename since I realized we're not uninstalling the plugin :-)

lib/tasks/foreman_templates_tasks.rake
ares: how about 'Clean default data seeded by this plugin, this results in permanent deletion!'

GregSutcliffe: Fair enough. I see this being run automatically in the future anyway, so lets not get dragged down over wording :)

GregSutcliffe:
Thanks @ares!

PR #45, opened 2017-04-13
Refs #19102 - add more protocols and tests

ares:


GregSutcliffe:
Works fine for me, generally ACK. This has come up a few times though - does it make sense to allow the user to override what things are counted as a repo? (similar to how we allow users to define what counts as an interface...)

ares:
could be a separate RFE, honestly I haven't seen such request yet, I think we now have all reasonable protocols :-)

GregSutcliffe:
Fair enough, if it comes up again we'll extend it.

Merging, thanks @ares!

PR #47, opened 2017-04-18
Adding button to sync foreman templates using a foreman task

sbernhard:
Adding a button to foreman_templates which does start the sync (import) of templates in a async foreman task.

### DIFF
--- a/app/controllers/templates_sync_controller.rb
+++ b/app/controllers/templates_sync_controller.rb
@@ -1,0 +1,13 @@
+class TemplatesSyncController < ApplicationController
+  def sync
+    begin
+      ForemanTasks.async_task(::Actions::TemplatesSync::Sync)
+      notice _("Added sync foreman templates task to queue, it will be run shortly.")
+
+    rescue ::Foreman::Exception => e
+      error _("Failed to add task to queue: %s") % e.to_s
+    ensure
+      redirect_to settings_path(:anchor => "TemplateSync")
+    end
+  end
+end
--- a/app/lib/actions/templates_sync/sync.rb
+++ b/app/lib/actions/templates_sync/sync.rb
@@ -1,0 +1,33 @@
+module Actions
+  module TemplatesSync
+    class Sync < Actions::Base
+
+      def plan
+        ::Foreman::Logging.logger('foreman_templates').info "Initiating 'sync' for templates."
+
+        # Make sure we don't have two concurrent listening services competing
+        if already_running?
+          fail "Action #{self.class.name} is already active"
+        end
+        plan_self
+      end
+
+      def run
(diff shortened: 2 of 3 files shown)


ares:
What a nice addition. This implies foreman_templates should depend on foreman_tasks which is quite a change :-) @GregSutcliffe or anyone else have any objection?

GregSutcliffe:
[test]

GregSutcliffe:
Thanks for the patch @sbernhard - great to have new features contributed :)

I'm a little wary of tying the plugin to foreman-tasks, when tasks itself is in the process of merging into core anyway. It feels like we might as well target this to release alongside 1.16 and make use of the ActiveJob async functionality there - would you agree?

sbernhard:
Sure. I wasn't aware of tasks will be merged to core. 

Would it be already possible to switch to ActiveJob async or do I need to wait till this job is finished?

ares:
There are no exact plans about foreman-tasks and core right now. An alternative way to do async tasks is ActiveJob API with dynflow backend, it's WIP, you can watch the progress at https://github.com/theforeman/foreman/pull/4316

After this gets in, it should be quite easy to use - http://api.rubyonrails.org/classes/ActiveJob/Base.html

sbernhard:
Thanks. If the ActiveJob API gets in to foreman 1.16, I wouldn't add the dependency to foreman-tasks within this commit. In case the ActiveJob API isn't ready for 1.16 I would suggest to put the PR in and later change it from foreman-tasks to ActiveJob API.

ares:
I agree. I expect that ActiveJob API will be available soon.

sbernhard:
Thanks. Using username: password would probably mean, that the password can be seen plain text in the UI. Wouldn't it be better to configure it separate as a setting? 


Von: "Marek Hulán" <notifications@github.com> 
An: "theforeman/foreman_templates" <foreman_templates@noreply.github.com> 
CC: "Bernhard Suttner" <bernhard.suttner@atix.de>, "Mention" <mention@noreply.github.com> 
Gesendet: Freitag, 21. April 2017 16:41:30 
Betreff: Re: [theforeman/foreman_templates] Adding button to sync foreman templates using a foreman task (#47) 



I agree. I expect that ActiveJob API will be available soon.

ares:
I think the best approach is to setup ssh key for the user under which the import runs.

ares:
ActiveJob work has been merged and can be now used. It has been demoed today. Would you be interested in looking into that? The recording should appear shortly on our youtube channel, let me know if you want a direct link.

ares:
I'm closing this for now for no activity, please reopen or let me know if you have a plan to continue working on this.

PR #52, opened 2017-05-17
Fixes #19555 - Add a setting to bypass template locks

xprazak2:
I created a new setting that allows bypassing the template lock, default value is false (do not update locked templates). It can be changed in the same way as any other setting for foreman_templates. 

I managed to improve messages a bit, users will know if template was skipped because it was locked. There is also something like:
```
template.errors if template.errors.any?
```
in the output, which is not very nice, but at least gives a hint what went wrong.

When executing import as non-admin user in an organization, user's org must be specified. Otherwise, there is  `Invalid organizations selection, you must select at least one of yours` error. When template has orgs in metadata that the user is not assigned to, they are ignored - template is associated only with user's organization. Same goes for locations.

When there is `OrgA` with template `some_tmpl` and `b_user` (user only in `OrgB`) tries to import `some_tmpl` (which is not yet present in `OrgB`), it failes with `name has already been taken`. This is a general problem with taxonomies - `some_tmpl` is not found for update in scope of `OrgB` and when we try to create it, we hit the unique constraint on name that is not scoped by taxonomy. 

I also managed to split the `import!` method a bit, but there is still a lot of duplicated code. I am not sure if further refactoring is worth the effort since we need to change the error handling and feedback to user.

~~TODO: tests~~

### DIFF
--- a/app/controllers/api/v2/template_controller.rb
+++ b/app/controllers/api/v2/template_controller.rb
@@ -9,6 +9,7 @@ module Api
       param :filter, String, :required => false, :desc => N_("Import names matching this regex (case-insensitive; snippets are not filtered).")
       param :negate, :bool, :required => false, :desc => N_("Negate the prefix (for purging).")
       param :associate, String, :required => false, :desc => N_("Associate to OS's, Locations & Organizations. Options are: always, new or never.")
+      param :force, :bool, :required => false, :desc => N_("Update templates that are locked")
       def import
         results = ForemanTemplates::TemplateImporter.new({
           verbose:   params['verbose'],
@@ -19,6 +20,7 @@ module Api
           filter:    params['filter'],
           associate: params['associate'],
           negate:    params['negate'],
+          force:     params['force']
         }).import!
         render :json => { :message => results }
       end
--- a/app/models/concerns/foreman_templates/provisioning_template_import.rb
+++ b/app/models/concerns/foreman_templates/provisioning_template_import.rb
@@ -3,9 +3,9 @@ module ForemanTemplates
     extend ActiveSupport::Concern
 
     module ClassMethods
-      def import!(name, text, metadata)
+      def import!(name, text, metadata, force = false)
         # Check for snippet type
-        return import_snippet!(name, text) if metadata['snippet'] || metadata['kind'] == 'snippet'
+        return import_snippet!(name, text, force) if metadata['snippet'] || metadata['kind'] == 'snippet'
 
         # Get template type
         kind = TemplateKind.find_by(name: metadata['kind'])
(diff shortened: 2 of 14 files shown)


app/controllers/api/v2/template_controller.rb
```diff
@@ -19,6 +19,7 @@ def import
           filter:    params['filter'],
           associate: params['associate'],
           negate:    params['negate'],
+          force:     params['force']
```
ares: I think this needs apipie documentation

app/models/concerns/foreman_templates/provisioning_template_import.rb
```diff
@@ -3,9 +3,9 @@ module ProvisioningTemplateImport
     extend ActiveSupport::Concern
 
     module ClassMethods
-      def import!(name, text, metadata)
+      def import!(name, text, metadata, force)
```
ares: I'm afraid that changing this API makes importing of JobTemplates incompatbile again. If that's the case, could you please also prepare a PR for rex (and openscap seed)? We could maybe make force argument optional, defaulting to false which would avoid the need for openscap seed change.

app/services/foreman_templates/template_importer.rb
```diff
@@ -6,15 +6,12 @@ class TemplateImporter < Action
...
+      super + %i(associate force)
     end
 
     def initialize(args = {})
       super
-      # Rake hands off strings, not booleans, and "false" is true...
-      if @verbose.is_a?(String)
-        @verbose = @verbose != 'false'
-      end
+      parse_bools('@verbose', '@force')
```
ares: I'm happy you extracted the logic to method but I'd prefer if you remained explicit about variables and avoid using `instance_variable_set` as it's hard to read, how about

```ruby
@verbose = parse_bool(@verbose)
@force  = parse_bool(@force)
```

xprazak2:
I added missing apipie doc, removed `instance_variable_set` and we are using default value for `force` in `import!` method

ares:
rubocop seems to be very sad :-(

xprazak2:
[test]

ares:
Works fine, local tests run seems OK.

```
Run options: --seed 22260

# Running:

............................................................................

Finished in 2.958844s, 25.6857 runs/s, 57.7928 assertions/s.

76 runs, 171 assertions, 0 failures, 0 errors, 0 skips
```

It would be great covering this in the plugin manual. Ideally the description should contain
* an example of output that user gets when template is locked
* explanation of force parameter
* how organizations/locations mapping works (similar to what you described here)
* ideally a new chapter "Common problems" or FAQ where we could add more common error messages, first candidate would be `name has already been taken` where the taxonomies issue would be explained.

The PR itself is ready for merge.

xprazak2:
[docs](https://github.com/theforeman/theforeman.org/pull/883)

ares:
Thanks @xprazak2, merging.

PR #54, opened 2017-05-19
Refs #19576 - Avoid using assert_equal for nil

xprazak2:
Jenkins is still red, this should fix the one remaining failure. We should avoid using `assert_equal` with `nil` in the future as it is now [deprecated](https://github.com/seattlerb/minitest/blob/master/lib/minitest/assertions.rb#L180-L183)


ares:
Tests running as http://ci.theforeman.org/job/test_plugin_pull_request/3710/

ares:
Seems like there's still some issue but this fixes some issues so I'll merge it. Thanks @xprazak2!

PR #58, opened 2017-07-19
Added API to README

lzap:
If you happen to have a curl example I will add it.


ares:
Thank @lzap, could you also mention export API? See https://theforeman.org/plugins/foreman_templates/5.0/index.html#5.API for curl examples. Maybe link from README there would be useful.

lzap:
Added new commit with the link.

PR #61, opened 2017-08-24
[WIP] Add rubocop-based validations task for templates

ShimShtein:
This PR adds a new rake task that will be able to validate provisioning templates to see if they comply to the latest template writing policy.
This should be extremely useful when foreman upgrade changes the way templates should be written. This could be used as an alternative to a migration, so the user will know what changes should be done to his custom templates.

## Usage

``` sh
rake templates:validate path/to/template/file.erb
```

As a developer, the only thing that should be done, is writing a new cop into `rubocop` folder.

## Algorithm
Currently the task works on a single source template file.
First it strips erb from any text that should not be evaluated as ruby and puts the result into a temporary file. It leaves every character at its original place, so when rubocop reports errors, they could easily be referenced in the original file (Done by `erb_strip.rb`).
Next rubocop is invoked to examine this file and put the result as JSON in another temp file.
Last step is cross-referencing rubocop results with the original template and outputting all offences in an appropriate format.

## TODO
 - [ ] Move rubocop invocation to task
 - [ ] Make list of all cops in `rubocop/` folder to add to rubocop
 - [ ] Create a config file instead of `--only` option.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,3 @@
 source "http://rubygems.org"
 
 gemspec
-gem 'rubocop'
--- a/app/services/foreman_templates/erb_strip.rb
+++ b/app/services/foreman_templates/erb_strip.rb
@@ -1,0 +1,31 @@
+module ForemanTemplates
+  class ErbStrip
+    ERB_REGEX = /(?<start_tag><%[-=]?)(?<script>([^#]).*?)(?<end_tag>-?%>)/m
+
+    def initialize(source)
+      @source = source
+    end
+
+    def strip
+      out = ''
+      current_position = 0
+      while (match = next_erb(current_position))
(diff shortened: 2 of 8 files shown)


GregSutcliffe:
See it in action here: https://www.youtube.com/watch?v=aOqA8-wpPKQ&feature=youtu.be&t=1029

@ShimShtein I assume you'll want to revisit this after the discussion on the demo?

ShimShtein:
Referencing here the mailing list thread that I will use for further offline discussion for those who are interested: https://groups.google.com/forum/#!topic/foreman-dev/KtVSqPKzXjA

ares:
@ShimShtein do you plan to continue working on this in near future?

PR #64, opened 2017-10-04
Support locking imported templates

treydock:
This would allow the behavior of this plugin to more closely mimic what's done by Foreman when it seeds in the community-templates repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ template will be automatically associated with the OS
 * dirname   => The directory within the git tree containing the templates [/]
 * filter    => Import names matching this regex (case-insensitive; snippets are not filtered)
 * associate => Associate to OS, "always", when "new" or "never"  [new]
+* lock      => Lock imported templates [false]
 
 The `branch` default will use *develop* if you're on Foreman-nightly; or the
 matching *1.X-stable* branch for your version of Foreman (if it exists); or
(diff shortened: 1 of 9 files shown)


xprazak2:
Hi, thanks for your patch. It works nicely, but there is a slight discrepancy between API and rake task.

1) create a git repo with 2 branches, make some changes between branches
2) import templates via rake task, in my case:
```
bin/rake templates:import repo=https://github.com/xprazak2/fake_templates.git branch=new-branch verbose=true lock=true force=false
```
result: it imports the templates and locks them

3) try importing from other branch:
```
bin/rake templates:import repo=https://github.com/xprazak2/fake_templates.git branch=develop verbose=true lock=true force=false
```
the templates get updated. I would expect them not to since they are locked and `force` is `false`.
If I do the same workflow through API, the locked templates are not updated.

Could you add the lock option into readme, please?

treydock:
@xprazak2 I'm not sure why the rake task would still be importing even when templates are locked.

My change does not yet handle snippets correctly, so this PR is likely incomplete.

I've updated the README.

treydock:
Updated so that lock is set for snippets but if a snippet's content does not differ but locking differs, the lock won't be applied.  Let me know if the snippet logic should be improved.

xprazak2:
I found out the cause why locked templates get unlocked via rake task and not via API. The culprit is the [`Template#locked?`](https://github.com/theforeman/foreman/blob/develop/app/models/template.rb#L26-L28) method. Easy fix is to replace it with just `locked` when importing.

Snippets should behave the same as other templates.

treydock:
@xprazak2 Changes made, snippets now behave like other template types and using `Template#locked` instead of `Template#locked?`.

xprazak2:
[test]

xprazak2:
Everything works fine now, I squashed the commits and merged as 3d7c198, thanks!

PR #67, opened 2017-11-01
Fix rubocop fix

ares:
@GregSutcliffe here I stand ashamed...


GregSutcliffe:
@ares we can both be ashamed, I didn't catch it :)

PR #68, opened 2017-11-03
Fixes rubocop after girl -> bot rename

ares:
as @xprazak2 found out elsewhere, indentation is off since "Girl".size == 4 while "Bot".size == 3 :-)


xprazak2:
That should do it, thanks!

PR #73, opened 2018-01-19
Fixes #22313 - Add purge into API

xprazak2:


xprazak2:
Closing as per comment in [redmine](https://projects.theforeman.org/issues/22313) - purge action was supposed to be removed and it can be replaced by bulk actions in core.

PR #77, opened 2018-01-22
Fixes #22374 - adapt to use importing from core

ares:
This illustrates how we could use importing from core in this plugin, https://github.com/theforeman/foreman/pull/5201 would need to be merged first


xprazak2:
Any plans moving ahead with this now that [core](https://github.com/theforeman/foreman/pull/5201) part has been merged or should I?

ares:
absolutely, just need to find a bit more time for this

ares:
Tests are broken by core changes, here's the output of local tests run

```
ares@edna:[~/z/foreman][develop✓]$ rake test:foreman_templates
Run options: --seed 64895

# Running:

...............................................

Finished in 2.534629s, 18.5431 runs/s, 44.5825 assertions/s.
47 runs, 113 assertions, 0 failures, 0 errors, 0 skips
```

ares:
Latest run with, locking requires https://github.com/theforeman/foreman/pull/5313 but does not block this PR, I can squash if you prefer.

```
Run options: --seed 47161

# Running:

...............................................

Finished in 3.029199s, 15.5157 runs/s, 37.3036 assertions/s.
47 runs, 113 assertions, 0 failures, 0 errors, 0 skips
```

app/services/foreman_templates/template_importer.rb
```diff
@@ -54,54 +51,52 @@ def parse_files!
...
+            template_type = metadata['model'].constantize
+          else
+            result_lines << "Metadata for template '#{name}' do not specify 'model', import failed"
+            next
+          end
+
+          template = template_type.import!(name, text, options)
+
+          data = {}
+          if template.template_changed?
```
xprazak2: Will always be `false`. `template_type.import!` on l.77 returns newly saved template, no diffs are ever shown.

xprazak2:
I also tested with the [locking](https://github.com/theforeman/foreman/pull/5313), encountered some unexpected behaviour. 

1) have an unlocked template
2) make change in template content
3) try to import with `'lock': 'true'`
4) you get an error that template is locked, but no changes to template content or lock are made

app/services/foreman_templates/template_importer.rb
```diff
@@ -54,54 +51,52 @@ def parse_files!
         result_lines << 'Parsing: ' + template.gsub(/#{@dir}#{@dirname}/, '') if @verbose
 
         metadata = parse_metadata(text)
```
xprazak2: Shouldn't we rather use `Template::parse_metadata`?

ares: :1st_place_medal:

app/services/foreman_templates/template_importer.rb
ares: bummer, `import!` seems to be too high-level, we need to hook into the process of saving to get the diff before save is called, that would actually help solving the problem with locking too, I'll look into it

ares:
> TemplateImporter#update_job_template is no longer used and can be removed, same for status_to_text.

Good catches!

> I also tested with the locking, encountered some unexpected behavior.

Did you also set `enforce: true`? Did the user have permission? Did you use rake task or API call? In general, we don't have control over bypassing the lock from this level as we don't have the instance yet. We could always pass enforce param, but then it kind of defeats the purpose. Note that user has to have permissions to lock the template - https://github.com/theforeman/foreman/blob/develop/app/models/template.rb#L182

Or was it this error? https://github.com/theforeman/foreman/blob/develop/app/models/template.rb#L193 I don't see how that would happen if we actually pass locked parameter, could you please check it bubbled there?

ares:
Updated, it seems to work with diff now, @xprazak2 please take another look.

xprazak2:
The diff shows now.

The locking validation error was indeed what [you suggested](https://github.com/theforeman/foreman/blob/develop/app/models/template.rb#L193). I was running as admin via API without `force`. The thing is that when you modify template text and try to lock it at the same time you have:

```ruby
actual_changes
#=> { :template => "whatever", :locked => [false, true] }

```

and the both parts of condition are true because we do not force it and `template` attribute is not in `allowed_changes`. But it will have to be fixed separately. Workaround is to do it in 2 steps - first change the template and then lock it.

I think you can squash and attach this to a [redmine issue](https://projects.theforeman.org/issues/22374)

ares:
After the change in core, I think this should now behave as expected. I squashed commits and linked it to the redmine issue, could you please take another look @xprazak2 ?

xprazak2:
Seems we are good here, the tests are now broken because of change in core, so this should actually fix them. Merging, thanks!

PR #78, opened 2018-01-23
_[wip] - Fixes #22374 - Deduplicate import logic

xprazak2:
Changes:
* I removed the import logic from template concers, importing is now handled by `TemplateParser`, so we have 1 method which imports all the templates
* templates need to have a couple of methods implemented, so that we know which associations to add, which attributes to import...
* message generation extracted to `ParseResult` - output remains the same as before, but it can now be changed more easily as it is contained within 1 class and separated from import logic.


xprazak2:
Closing in favour of #77

PR #79, opened 2018-01-23
Fixes #22380 - Have prefix empty by default

xprazak2:


ares:
@jyejare I guess this might be a good candidate for testing :-)

jyejare:
Tested, and It works as expected.

Tests Performed:
```
1. Test: Prefix setting should have by default 'Empty' in Settings-> TemplateSync in UI
Assert: Prefix settings have 'Empty'.

2. Test: Importing templates using rake templates:sync should set no prefix to imported templates.
Assert: Imported templates have no prefix.

3. Test: Change prefix setting in Settings-> TemplateSync to some value, and check if importing templates are prefixed with that value.
Assert: Changing the prefix setting worked and imported templates has prefixed value.
```

ares:
Thanks @xprazak2, code looks also good, I retriggered the jenkins master job, good to be merged when tests are green.

ares:
Thanks @xprazak2, merged!

PR #80, opened 2018-01-25
Fixes #22353 - Remove prefix from template name when exporting

xprazak2:
Adds a new setting that allows removing prefix when exporting, therefore template name does not have multiple prefixes when running export/import repeatedly. Default value is set to false.

### DIFF
--- a/app/services/foreman_templates/template_exporter.rb
+++ b/app/services/foreman_templates/template_exporter.rb
@@ -62,6 +62,8 @@ module ForemanTemplates
         current_dir = get_dump_dir(template)
         FileUtils.mkdir_p current_dir
 
+        set_name_without_prefix(template)
+
         filename = File.join(current_dir, get_template_filename(template))
         File.open(filename, 'w+') do |file|
           logger.debug "Writing to file #{filename}"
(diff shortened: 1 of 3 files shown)


ares:
just quickly read the PR, wouldn't it be better if exporting checks whether prefix is already part of the name and in that case it would ignore it? I think it already worked that way. I think adding extra setting makes it even more complicated to understand the prefix logic.

xprazak2:
Rebased and updated. We now check if the name is prefixed and if so, prefix is ignored.

xprazak2:
[test]

ares:
I'm unable to reproduce, exporting does never add prefix (and it seems it never did). Import adds prefix but export ignores it. The prefix parameter is not even documented in export API endpoint. The confusing might be the negate setting description, which says "Negate the prefix (for purging) / filter (for importing/exporting)". `/exporting` should be dropped from it I think and that should be all.

xprazak2:
The setting description is correct, we use [filter](https://github.com/theforeman/foreman_templates/blob/master/app/services/foreman_templates/template_exporter.rb#L87) for import/export but prefix for purging.

I know this issue was present, but I cannot reproduce any more either. I guess it got fixed with recent changes. Closing, will reopen if it pops up again.

PR #81, opened 2018-02-09
Fixes #22538 - 'repo' should mention filesystem too

dLobatog:
The "Repo" setting can be used to export/import from a filesystem,
ssh, etc..
The documentation and the repository README explain that if it
points to '/tmp/' (just an example) it will export templates to
that destination. I tested it and it works well.
However, on Administer > Settings, the text explicitly says "Git
repo", which may confuse an user into thinking they can only use a
Git repo as the target.

### DIFF
--- a/app/models/setting/template_sync.rb
+++ b/app/models/setting/template_sync.rb
@@ -28,7 +28,7 @@ class Setting
           self.set('template_sync_prefix', N_('The string all imported templates should begin with'), "Community ", N_('Prefix')),
           self.set('template_sync_dirname', N_('The directory within the Git repo containing the templates'), '/', N_('Dirname')),
           self.set('template_sync_filter', N_('Import or export names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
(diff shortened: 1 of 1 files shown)


app/models/setting/template_sync.rb
```diff
@@ -28,7 +28,7 @@ def self.load_defaults
           self.set('template_sync_prefix', N_('The string all imported templates should begin with'), "Community ", N_('Prefix')),
           self.set('template_sync_dirname', N_('The directory within the Git repo containing the templates'), '/', N_('Dirname')),
           self.set('template_sync_filter', N_('Import or export names matching this regex (case-insensitive; snippets are not filtered)'), nil, N_('Filter')),
-          self.set('template_sync_repo', N_('Default Git repo to sync from'), 'https://github.com/theforeman/community-templates.git', N_('Repo')),
+          self.set('template_sync_repo', N_('Target path to import and export. Different protocols can be used, e.g. /tmp/dir or git://example.com, https://example.com, ssh://example.com'), 'https://github.com/theforeman/community-templates.git', N_('Repo')),
```
sean797: Mind moving the "or" to the last example OR replacing it with a comma ?

xprazak2:
Test failure unrelated, works as expected. Thanks!

PR #89, opened 2018-04-11
Fixes #19088 - Add params for taxonomies on import/export

xprazak2:


jyejare:
Tests Performed:
```
1. Import - Template can be imported in specific ORG and LOC.
2. Import - Template can be imported in specific ORG.
3. Import - Template can be imported in specific LOC.
4. Export - Template can be exported from specific ORG and LOC
5. Export - Template can be exported from specific ORG.
6. Export - Template can be exported from specific LOC.
```

All the tests above are passing.


The only thing I observed is APIDOC is yet to be updated for export and import to use:

1. **organization_ids** instead of **organization_id**, and **list** instead of **integer**.
2. **location_ids** instead of **location_id**, and **list** instead of **integer**.

app/services/foreman_templates/template_exporter.rb
```diff
@@ -107,5 +107,41 @@ def export_method
...
+      end
+    end
+
+    def templates_query(template_class, tax_ids)
+      template_class.joins(:taxable_taxonomies).where(:taxable_taxonomies => { :taxonomy_id => tax_ids })
+    end
+
+    def taxes_ids(tax_type)
+      tax_type.capitalize.constantize
+              .where(:name => instance_variable_get("@#{tax_type.pluralize}")["#{tax_type}_names".to_sym])
```
ares: please define attr_accessor to avoid using instance_variable_get, public_send is more readable or even better would be something like `@parameters[tax_type]['names']` that would be created in action initializer, to avoid this dynamic programming and be less dependent on parameters structure.

app/services/foreman_templates/template_exporter.rb
```diff
@@ -107,5 +107,41 @@ def export_method
           raise "Unknown metadata export mode #{@metadata_export_mode}"
       end
     end
+
+    private
+
+    def find_templates
+      @organizations.empty? && @locations.empty? ? Template.all : find_taxed_templates
```
ares: `(@organizations + @locations).empty?`  is slightly less repetitive

app/services/foreman_templates/template_exporter.rb
```diff
@@ -107,5 +107,41 @@ def export_method
...
+
+    def find_taxed_templates
+      Template.subclasses.reduce([]) do |memo, subclass|
+        memo.concat templates_in_tax(subclass)
+      end
+    end
+
+    def templates_in_tax(template_class)
+      location_ids = taxes_ids('location')
+      organization_ids = taxes_ids('organization')
```
ares: you can keep ids in a single array, their ids will always be unique, e.g. `taxonomy_ids = location_ids + organization_ids`

then below it does not really matter what model do you use, you could just do
```
templates_query(taxable_ids)
```
where templates_query would be using `Taxonomy` parent directly, e.g.
```
Taxonomy.join(:taxable_taxonomies).where(:taxable_taxonomies => { :taxonomy_id => tax_ids })
```

Rails will correctly load the type when it builds instances.

xprazak2:
Changes: I store all taxonomy params in a hash and avoid accessing the instance variables dynamically. I also improved the template query a bit, but keeping both location and organization ids in one array does not work, because we want templates that are in both, which means intersection of the two rather than union.

test/unit/template_exporter_test.rb
```diff
@@ -67,10 +101,10 @@ class TemplateExporterTest < ActiveSupport::TestCase
         assert_equal "/tmp/provisioning_templates/snippet", @exporter.get_dump_dir(@template)
       end
 
-      test 'does not appends anything for template without kind' do
+      test 'appends _templates for ptables' do
         @exporter.instance_variable_set('@dir', '/tmp')
         template = FactoryBot.create(:ptable)
-        assert_equal "/tmp/ptables/", @exporter.get_dump_dir(template)
+        assert_equal "/tmp/partition_tables_templates/", @exporter.get_dump_dir(template)
```
ares: this does not seem to work, export dir is still ptables, it would also become incompatible with community-templates repo, is the change intentional? I think it's ready for merge after reverting this

ares:
Thanks @xprazak2, merging!

PR #92, opened 2018-05-11
Fixes #23570 - import metadata separately

ares:


ares:
requires https://github.com/theforeman/foreman/pull/5545 in core to be merged first, then it should bump dependency probably to Foreman 1.19

PR #95, opened 2018-06-20
Fixes #23999 - Respect current context when exporting

xprazak2:


ares:
actually not merging as there new cop offense

```
/var/lib/workspace/workspace/foreman_templates-pull-request/database/postgresql/label/fast/ruby/2.5/plugin/app/services/foreman_templates/template_exporter.rb:136:36: C: Style/SymbolProc: Pass &:all as an argument to flat_map instead of a block.
      Template.subclasses.flat_map do |subclass| ...
                                   ^^^^^^^^^^^^^
```

PR #99, opened 2018-06-28
Fixes #24089 - Add more flexibility to lock setting

xprazak2:
needs https://github.com/theforeman/foreman/pull/5752

### DIFF
--- a/app/controllers/concerns/foreman/controller/parameters/template_params.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/template_params.rb
@@ -39,16 +39,22 @@ module Foreman
 
         def transform_lock_param(params)
           lock = params[:lock]
-          return params unless lock
+          return params if lock.nil?
 
-          if lock == "true" || lock.is_a?(TrueClass)
+          if lock == "true" || lock.is_a?(TrueClass) || lock.to_s == "1"
+            log_deprecated_param(lock)
             params[:lock] = "lock"
           end
 
-          if lock == "false" || lock.is_a?(FalseClass)
+          if lock == "false" || lock.is_a?(FalseClass) || lock.to_s == "0"
+            log_deprecated_param(lock)
             params[:lock] = "unlock"
           end
           params
+        end
+
+        def log_deprecated_param(value)
+          Logging.logger('app').warn "Using '#{value}' as a value for lock when syncing templates is deprecated and will be removed in the future."
         end
 
         def template_export_params
--- a/app/controllers/concerns/foreman/controller/parameters/template_params.rb
+++ b/app/controllers/concerns/foreman/controller/parameters/template_params.rb
@@ -34,7 +34,21 @@ module Foreman
         end
 
         def template_import_params
-          add_taxonomy_params(base_import_params(:none))
+          transform_lock_param add_taxonomy_params(base_import_params(:none))
+        end
+
+        def transform_lock_param(params)
+          lock = params[:lock]
+          return params unless lock
+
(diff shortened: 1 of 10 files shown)


xprazak2:
Rebased

bkearney:
@xprazak2 this is a pretty old bug, can you poke someone to get it reviewed and merged?

app/services/foreman_templates/template_importer.rb
```diff
@@ -79,11 +78,32 @@ def parse_files!
...
+      lock_predicate = lambda do |template|
+        case @lock
+        when 'lock'
+          return true
+        when 'unlock'
+          return false
+        when 'keep'
+          return template.locked unless template.new_record?
+
+          return false
```
ofedoren: Would it do the same if we have `template.new_record? ? false : template.locked` ?

app/services/foreman_templates/template_importer.rb
```diff
@@ -79,11 +78,32 @@ def parse_files!
...
+        when 'keep_lock_new'
+          return template.locked unless template.new_record?
+
+          return true
```
ofedoren: dtto: `template.new_record? ? true : template.locked` ?

xprazak2:
I made changes to score the extra points for backward compatibility.

app/controllers/concerns/foreman/controller/parameters/template_params.rb
```diff
@@ -34,7 +34,21 @@ def ui_template_export_params
         end
 
         def template_import_params
-          add_taxonomy_params(base_import_params(:none))
+          transform_lock_param add_taxonomy_params(base_import_params(:none))
+        end
+
+        def transform_lock_param(params)
+          lock = params[:lock]
+          return params unless lock
```
ofedoren: If we support both string `"true"`/`"false"` and boolean `true`/`false` (as I see below) then probably we should change it to `return params if lock.nil?`, shouldn't we?

xprazak2: Good catch.

xprazak2:
I also added '0' and '1' + a warning into logs.

app/controllers/concerns/foreman/controller/parameters/template_params.rb
```diff
@@ -34,7 +34,27 @@ def ui_template_export_params
...
         def template_import_params
-          add_taxonomy_params(base_import_params(:none))
+          transform_lock_param add_taxonomy_params(base_import_params(:none))
+        end
+
+        def transform_lock_param(params)
+          lock = params[:lock]
+          return params if lock.nil?
+
+          if lock == "true" || lock.is_a?(TrueClass) || lock == "1"
```
ofedoren: This is okay, but wouldn't be safer to compare strings? E.g. `lock.to_s == '1'` in case if lock is integer?

app/controllers/concerns/foreman/controller/parameters/template_params.rb
```diff
@@ -34,7 +34,27 @@ def ui_template_export_params
...
+            log_deprecated_param(lock)
+            params[:lock] = "lock"
+          end
+
+          if lock == "false" || lock.is_a?(FalseClass) || lock == "0"
```
ofedoren: Dtto lock.to_s == '0'

xprazak2:
Yes, it would be better.

ares:
Thanks @xprazak2 and @ofedoren, merging!

PR #104, opened 2018-08-16
Foreman Templates Tier1 API tests automated

jyejare:
Automated Tier1(Import and Export) tests.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -84,6 +84,9 @@ Style/WordArray:
 Rails/Output:
   Enabled: false
 
+Style/ClassAndModuleChildren:
+  Enabled: false
+
 # needs major refactorings before we can improve
 Metrics/ClassLength:
   Max: 200
--- a/test/functional/api/v2/template_controller_test.rb
+++ b/test/functional/api/v2/template_controller_test.rb
@@ -3,9 +3,13 @@ require 'test_plugin_helper'
 module Api
   module V2
     class TemplateControllerTest < ::ActionController::TestCase
+      setup do
+        @org = FactoryBot.create(:organization, :name => 'ForemanTemplatesOrg')
+        @loc = FactoryBot.create(:location, :name => 'ForemanTemplatesLoc')
+      end
 
       test "should import from git" do
-        post :import, params: { 'repo' => "#{ForemanTemplates::Engine.root}/test/templates/core", 'prefix' => '' }
+        post :import, params: { 'repo' => "#{template_fixtures_path}/core", 'prefix' => '' }
         assert_response :success
       end
 
@@ -27,7 +31,7 @@ module Api
       end
 
       test "should import filtered templates" do
-        post :import, params: { "repo" => "#{ForemanTemplates::Engine.root}/test/templates/core", "filter" => 'filter1', "prefix" => "filtTemp_", "associate" => "new"}
+        post :import, params: { "repo" => "#{template_fixtures_path}/core", "filter" => 'filter1', "prefix" => "filtTemp_", "associate" => "new" }
         assert_response :success
         templates = JSON.parse(@response.body)['message']['templates']
         templates.each do |template|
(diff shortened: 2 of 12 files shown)


test/functional/api/v2/template_controller_test.rb
```diff
@@ -1,8 +1,30 @@
 require 'test_plugin_helper'
+require 'git'
 
 module Api
   module V2
     class TemplateControllerTest < ::ActionController::TestCase
+
+      private
+      def create_local_template_in_core(t_filename:, t_name:, model: "ProvisioningTemplate", kind: "provision", org: "", loc: "")
```
xprazak2: It has no impact on tests, but it is customary to put private methods at the end, because the `private` makes all methods below it private:

```ruby
2.3.4 :010 > class Cow
2.3.4 :011?>   private
2.3.4 :012?>   def crazy_moo
2.3.4 :013?>     "mooooah"
2.3.4 :014?>     end
2.3.4 :015?>   def moo
2.3.4 :016?>     "moo"
2.3.4 :017?>    end
2.3.4 :018?>  end
 => :moo 
2.3.4 :019 > c = Cow.new
 => #<Cow:0x000000025602c8> 
2.3.4 :020 > c.crazy_moo
NoMethodError: private method `crazy_moo' called for #<Cow:0x000000025602c8>
	from (irb):20
	from /home/vagrant/.rvm/rubies/ruby-2.3.4/bin/irb:11:in `<main>'
2.3.4 :021 > c.moo
NoMethodError: private method `moo' called for #<Cow:0x000000025602c8>
	from (irb):21
	from /home/vagrant/.rvm/rubies/ruby-2.3.4/bin/irb:11:in `<main>'
2.3.4 :022 >
```

You could use `public` to make methods public again:

```ruby
2.3.4 :001 > class Cow
2.3.4 :002?>   private
2.3.4 :003?>   def crazy_moo
2.3.4 :004?>     "moooah"
2.3.4 :005?>     end
2.3.4 :006?>   public
2.3.4 :007?>   def moo
2.3.4 :008?>     "mooo"
2.3.4 :009?>    end
2.3.4 :010?>  end
 => :moo 
2.3.4 :011 > c = Cow.new
 => #<Cow:0x0000000147c358> 
2.3.4 :012 > c.crazy_moo
NoMethodError: private method `crazy_moo' called for #<Cow:0x0000000147c358>
	from (irb):12
	from /home/vagrant/.rvm/rubies/ruby-2.3.4/bin/irb:11:in `<main>'
2.3.4 :013 > c.moo
 => "mooo"
```

test/functional/api/v2/template_controller_test.rb
```diff
@@ -1,8 +1,30 @@
...
+        File.open("#{ForemanTemplates::Engine.root}/test/templates/core/#{t_filename}.erb", "w") { |template|
```
xprazak2: Note that this creates a file in the static fixtures folder. If not used carefully, it can overwrite existing files thus breaking other tests that rely on them, so I think we should use a different path. In the ideal scenario we would have the template as a temporary file only for the duration of the test.

test/functional/api/v2/template_controller_test.rb
```diff
@@ -1,8 +1,30 @@
...
+          template << "<%#\n"
+          template << "model: #{model}\n"
+          template << "kind: #{kind}\n"
+          template << "name: #{t_name}\n"
+          template << "oses:\n"
+          template << "- Debian 6.0\n"
+          if org
+            template << "organizations: #{org}\n"
```
xprazak2: You can do :

```ruby
template << "organizations: #{org}\n" if org
```

test/functional/api/v2/template_controller_test.rb
```diff
@@ -1,8 +1,30 @@
...
+          end
+          if loc
+            template << "locations: #{loc}\n"
```
xprazak2: same as above

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +46,128 @@ class TemplateControllerTest < ::ActionController::TestCase
...
+        assert_response :success
+        updated = JSON.parse(@response.body)['message']['templates']
+        updated.each do |template|
+          if template['name'] == 'filter1 template'
+            puts template
+          end
+        end
+      end
+
+      test "should template associates to taxonomies in metadata on import" do
```
xprazak2: Maybe "should associate template to taxonomies in metadata on import"?

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +46,128 @@ class TemplateControllerTest < ::ActionController::TestCase
...
+        tname = "metadata_with_taxonomies"
+        org = taxonomies(:organization1)
+        loc = taxonomies(:location1)
+        create_local_template_in_core t_filename: "metadata-1.erb", t_name: tname, org: org.name, loc: loc.name
```
xprazak2: Nit: inconsistent underscores - `t_name` vs `tname`

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +46,128 @@ class TemplateControllerTest < ::ActionController::TestCase
...
+        assert_response :success
+        templates = JSON.parse(@response.body)['message']['templates']
+        assert_equal templates.length, 10
+        templates.each do |template|
+          assert template['imported']
+        end
+      end
+
+      test "should not update locked template on import" do
+        FactoryBot.create(:import, :repo => "#{ForemanTemplates::Engine.root}/test/templates/core", :filter => "filter1", :prefix => '', :associate => "new", :lock => true)
```
xprazak2: Test are failing on this, maybe it should be `post :import` ?

jyejare: Then there will be two posts in the same test for which we are advised to not to do so. What do u think ?

jyejare: Also, I thought its failing for my system only but fortunately on ur s as well.

xprazak2: There is no factory for `:import`, if you do not want to use a second `post`, then you will need to read the template files and create the templates that you need.

jyejare: I think, better I will use two posts.

xprazak2:
[test]

jyejare:
@xprazak2 :

```
Api::V2::TemplateControllerTest.test_0011_should export filtered template only
```
This Test is actually failing because the export folder **xyz** is being created instead of **ProvisioningTemplates/Provision**.

```
Api::V2::TemplateControllerTest.test_0007_should import from git repository directory
Api::V2::TemplateControllerTest.test_0004_should import filtered templates
Api::V2::TemplateControllerTest.test_0006_should import from git branch
Api::V2::TemplateControllerTest.test_0012_should import accordingly to exported template metadata
```
These four are fixed, lets see what test says.

jyejare:
[test]

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +25,146 @@ class TemplateControllerTest < ::ActionController::TestCase
...
+          assert Dir.entries("#{tmpdir}/provisioning_templates/provision").include?('exporting_template.erb')
+        end
+      end
+
+      test "should import accordingly to exported template metadata" do
+        org = taxonomies(:organization2)
+        loc = taxonomies(:location2)
+        name = "taxo_metadata_import"
+        templte = "<%#\nmodel: ProvisioningTemplate\nkind: provision\nname: #{name}\norganizations: #{org.name}\nlocations: #{loc.name}\n-%>\nimport metadata"
+        ProvisioningTemplate.create!(:name => name, :template => templte)
```
xprazak2: You need to specify `template_kind` for the template to pass the validation.

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +25,146 @@ class TemplateControllerTest < ::ActionController::TestCase
...
+        Dir.mktmpdir do |tmpdir|
+          FactoryBot.create(:export, :repo => tmpdir, :filter => name, :metadata_export_mode => "refresh")
```
xprazak2: This will not work for the same reasons as `FactoryBot.create(:import, ...)`. If you want to test that repeated imports and exports are idempotent, you will need to call `post` for each of them. If you only want to test that import is correct based on what is in metadata, you may need to prepare the files in `tmpdir` prior to doing import.

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +25,146 @@ class TemplateControllerTest < ::ActionController::TestCase
...
+          post :import, params: { "repo" => tmpdir, "filter" => t_name, "prefix" => '', "associate" => "always" }
+          assert_response :success
+          assert_equal ProvisioningTemplate.unscoped.find_by_name(t_name).organizations[0].name, org.name
+          assert_equal ProvisioningTemplate.unscoped.find_by_name(t_name).locations[0].name, loc.name
+        end
+      end
+
+      test "should export filtered template only" do
+        Dir.mktmpdir do |tmpdir|
+          FactoryBot.create(:provisioning_template, :name => 'exporting_template')
```
xprazak2: Try:
```ruby
kind = TemplateKind.find_by :name => 'provision'
FactoryBot.create(:provisioning_template, :name => 'exporting_template', :template_kind => kind)
```

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +25,146 @@ class TemplateControllerTest < ::ActionController::TestCase
           refute Dir.entries("#{tmpdir}/provisioning_templates/provision").include?('export_test_template.erb')
         end
       end
+
+      test "should import filtered templates" do
+        post :import, params: { "repo" => "#{ForemanTemplates::Engine.root}/test/templates/core", "filter" => 'filter1', "prefix" => "filtTemp_" }
+        assert_response :success
+        templates = JSON.parse(@response.body)['message']['templates']
+        templates.each do |template|
+          if template['name'] == 'filtTemp_filter1 template'
```
xprazak2: `filtTemp_filter1 template` was not imported due to missing template kind.

jyejare: But, It is there mentioned already in filter1 template -> https://github.com/theforeman/foreman_templates/pull/104/files#diff-f78c60bb7753401d1ec4a66c7b2630d3R2. It should be a bug.

xprazak2: Not really. You need to specify you want to create associations based on [metadata](https://github.com/theforeman/foreman/blob/develop/app/models/template.rb#L215)

```ruby
post :import, params: { "repo" => "#{ForemanTemplates::Engine.root}/test/templates/core", 
                                    "filter" => 'filter1',
                                    "prefix" => "filtTemp_",
                                     "associate" => "new" }
```

test/functional/api/v2/template_controller_test.rb
```diff
@@ -24,6 +25,146 @@ class TemplateControllerTest < ::ActionController::TestCase
...
+          end
+        end
+      end
+
+      test "should import from git branch" do
+        post :import, params: { "repo" => 'https://github.com/SatelliteQE/foreman_templates.git', "prefix" => "gitBranch", "branch" => "example", "associate" => "new" }
+        assert_response :success
+        templates = JSON.parse(@response.body)['message']['templates']
+        assert_equal templates.length, 11
+        templates.each do |template|
```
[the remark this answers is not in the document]
jyejare: But, as I checked in gitrepo 'example' branch, the template kind is mentioned on all the templates. Earlier this test used to pass. but not sure whats happenning now.

test/functional/api/v2/template_controller_test.rb
xprazak2: In the template string, yes. But `create!` method does not parse the template string. It just creates a record with given attributes and provisioning template needs template kind association to be [valid](https://github.com/theforeman/foreman/blob/develop/app/models/provisioning_template.rb#L22), therefore you need to have

```ruby
ProvisioningTemplate.create!(:name => name, :template => template, :template_kind => template_kinds(:provision))
```
for validation to pass.

It turns out `ActionController::TestCase` supports only testing [one action](https://api.rubyonrails.org/classes/ActionController/TestCase.html). A workaround might be to call the exporter [directly](https://github.com/theforeman/foreman_templates/blob/master/app/controllers/api/v2/template_controller.rb#L35) or create [integration test](https://api.rubyonrails.org/classes/ActionDispatch/IntegrationTest.html) if you need both requests to happen.

test/functional/api/v2/template_controller_test.rb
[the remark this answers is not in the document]
xprazak2: The problem is that `user_data` kind is not in core's [fixtures](https://github.com/theforeman/foreman/blob/develop/test/fixtures/template_kinds.yml) so you will need to create it prior to importing.

xprazak2:
@jyejare, any progress on this?

jyejare:
Can close this or  @xprazak2 could you take this up ?

xprazak2:
You can keep it open, I'll take over when I have a spare moment, tests are always good to have.

jyejare:
@xprazak2 Sure, Thanks for taking care of it !

xprazak2:
I pushed a new commit to make the tests pass. Main changes:

* using file fixtures rather that creating template files dynamically inside tests

* I removed the tests that rely on external github repo, we might add them [later](https://projects.theforeman.org/issues/28466)

* using integration test for complicated workflows

.rubocop.yml
```diff
@@ -84,6 +84,12 @@ Style/WordArray:
 Rails/Output:
   Enabled: false
 
+Style/ClassAndModuleChildren:
+  Enabled: false
+
+Rails/HttpPositionalArguments:
```
mmoll: Rails 5.x+ needs this, what's the background of having it disabled here?

ares:
Thanks @jyejare and @xprazak2, merging!

jyejare:
@xprazak2 and @ares Thanks for taking this up and merging !!

PR #105, opened 2018-08-31
Fixes #24627 - Unclear hammer error message while sync

ofedoren:
Since the error was under :message => { :error => "error" } hammer tried to print the hash instead of the error message.

I'm not actually sure if this is a hammer-cli core problem or the API response inconsistency.

### DIFF
--- a/app/controllers/api/v2/template_controller.rb
+++ b/app/controllers/api/v2/template_controller.rb
@@ -33,7 +33,12 @@ module Api
       param_group :taxonomies, ::Api::V2::BaseController
       def export
         @result = ForemanTemplates::TemplateExporter.new(template_export_params).export!
-        render :json => { :message => @result.to_h }, :status => @result.exported ? 200 : 500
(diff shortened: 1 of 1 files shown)


xprazak2:
Thanks!

xprazak2:
@ofedoren, any chance this could be fixed in hammer-cli core somehow? Hammer prints a nice message now, but `:message` is an object when response is a success or a string when there is an error. I kind of liked that we had the same structure before.

ofedoren:
@xprazak2, I figured out that I can [easily solve](https://github.com/theforeman/hammer-cli-foreman-templates/commit/fa22ddbe6681f4f52bfd8f27c523d4f2bcb2e992) this in the hammer plugin, so there is no need to change the API response.

And I apologize for troubling you with reverting of this PR.

xprazak2:
Thanks, the fix in hammer works, I reverted here.

PR #111, opened 2019-10-04
Fixes #27967 - Change back link into button

xprazak2:


ares:
The button ends up with "active" css class and looks like pressed on my environment, can you also reproduce that?

PR #112, opened 2019-10-04
Fixes #27963 - Fix name column on sync result

xprazak2:


ares:
Thanks @xprazak2, failures are unrelated, merging!

PR #119, opened 2019-10-10
Refs #28025 - fix pageLayout mocks

amirfefer:
to keep consistency with the `pageLayout` change


ares:
Thanks @amirfefer, merging

PR #121, opened 2019-10-15
Fixes #28061 - Update npm dependencies

xprazak2:
This is probably post 1.24, so I would wait with merging.


xprazak2:
We need to update foreman/vendor to make [koji](http://koji.katello.org/kojifiles/work/tasks/7833/237833/mock_output.log) happy.

PR #125, opened 2019-11-13
Fixes #28256 - upgrade vendor to v3

sharvit:
Main upgrades are `redux-hooks` and `babel` v7


xprazak2:
Looks good, I only found out the form fields do not show with the new vendor: 

missing-fields


Moving from redux-forms to Formik helps and we want to do it anyway. Could you please take a look at https://github.com/theforeman/foreman_templates/pull/131 ?

sharvit:
> I only found out the form fields do not show with the new vendor

It is very odd, wondering if it caused by one of the vendor updates or by one of those PRs:
- https://github.com/theforeman/foreman/pull/6832
- https://github.com/theforeman/foreman/pull/7189

Anyway, I would love to fix it so forms using `redux-forms` would keep working :worried: 



> Moving from redux-forms to Formik helps and we want to do it anyway. Could you please take a look at #131 ?

Will do! thanks!

sharvit:
Updated to support vendor v4 as well

xprazak2:
Travis failures related, seems like some linting rules have changed.

sharvit:
Thanks @xprazak2, fixed.

xprazak2:
This might need an additional change, Travis is still red.

sharvit:
Thanks @xprazak2, it works now, it was probably a cache problem with travis and npm.
Anyway, I updated `eslint`, `eslint-patternfly-react` and `prettier`. Also, removed some unnecessary eslint plugins.

xprazak2:
Thanks!

PR #128, opened 2019-12-04
Fixes #28414 - Show file name on template export

xprazak2:


app/models/concerns/foreman_templates/template_extensions.rb
```diff
@@ -0,0 +1,11 @@
+require 'shellwords'
+
+module ForemanTemplates
+  module TemplateExtensions
+    extend ActiveSupport::Concern
+
+    def template_file
+      Shellwords.escape(name.downcase.tr(' /', '_') + '.erb')
```
adamruzicka: Is the escaping necessary?

lib/foreman_templates/engine.rb
```diff
@@ -50,5 +50,9 @@ class Engine < ::Rails::Engine
              :turbolinks => false
       end
     end
+
+    config.to_prepare do
+      Template.send(:include, ForemanTemplates::TemplateExtensions)
```
adamruzicka: The send shouldn't be necessary

app/models/concerns/foreman_templates/template_extensions.rb
xprazak2: I think it might be needed:

```ruby
"a/b'c d".downcase.tr(' /', '_')
#=> "a_b'c_d"

Shellwords.escape "a/b'c d".downcase.tr(' /', '_')
#=> "a_b\\'c_d"
```

adamruzicka: Does it matter that there's a quote in the name? Afaik we shouldn't be passing the name to a shell anywhere

ares:
seems like tests in core are broken, perhaps https://github.com/theforeman/foreman/pull/7231

that prevents plugin tests to run, we'll probably need to wait or help reviewing that PR

adamruzicka:
The core PR was merged, let's [test foreman_templates]

ares:
all green, any more comments @adamruzicka ?

adamruzicka:
Nope, looks good to me

ares:
Thanks @xprazak2 and @adamruzicka, merging!

PR #131, opened 2019-12-12
Fixes # 28492 - Use formik instead of redux-forms

xprazak2:


xprazak2:
[test foreman_templates]

sharvit:
Thanks @xprazak2 

IMO the form is too complicated because there is one form that **acts** as 2 different forms and I would like to suggest some refactoring.

1. Split the form into 2 forms. ImportForm and ExportForm.
2. Use tabs instead of radio-buttons and switch them by react-router.

Thoughts?

xprazak2:
Mapping each form to a separate route provided by react-router would be nice, I opened [a new redmine issue](https://projects.theforeman.org/issues/28615) for that and refactor after we swap the form framework.

webpack/components/NewTemplateSync/components/NewTemplateSyncForm/__tests__/__snapshots__/NewTemplateSyncForm.test.js.snap
```diff
@@ -1,11 +1,116 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`NewTemplateSyncForm should render for export settings 1`] = `
-<Form
-  disabled={false}
-  errorTitle="Warning! "
+<ForemanForm
```
sharvit: IMO this test/snapshot falls into the definition of a **False positive** and **False negative**.

Most of the data in the snapshot look irrelevant while the only relevant field is the `initialValues`.

webpack/components/NewTemplateSync/components/SyncSettingFields.js
```diff
@@ -58,14 +42,15 @@ const SyncSettingsFields = ({
   return (
     <React.Fragment>
       {settingsAry
-        .map(setting => addValidationToSetting(setting, validationData))
+        .map(addRequiredToSetting)
```
sharvit: The way it works is a bit odd, the `SyncSettingsFields` receives `importSettings`, `exportSettings` and `syncType`.
Based on the `syncType` it will use `importSettings` or `exportSettings`.

IMO it should be done at a higher level so this component would only receive the final `settingsAry`.
Same for the manipulations done on the `settingsAry`, it can all be done in a **memorized selector**.

Please consider refactoring, maybe it should be done in a different PR.

xprazak2:
Those are very good points @sharvit! I'd be definitely willing to improve the structure but I'd prefer to do it in a separate PR to make it obvious which changes were maintenance refactoring and what was needed for formik.

sharvit:
> Those are very good points @sharvit! I'd be definitely willing to improve the structure but I'd prefer to do it in a separate PR to make it obvious which changes were maintenance refactoring and what was needed for formik.

Make sense, thanks.
My only concern is the test snapshots, it looks like things might break but the tests won't catch it (False positive). And from the other side, the test may produce errors while the app would keep functioning (False negative).

False-positive might happen because the snapshot only validates the `initialFields`.
False-negative might happen because the snapshot contains the `validationSchema` that might be changed when we will change/update the validations library.

Is there anything we can do about it before the refactoring?

xprazak2:
[test foreman_templates]

xprazak2:
@ares, can we merge?

PR #134, opened 2020-01-10
Fixes #28716 - Create correctly branch for empty repo

xprazak2:


app/services/foreman_templates/.fuse_hidden0000028600000002
```diff
@@ -0,0 +1,163 @@
+module ForemanTemplates
```
ares: this shouldn't probably be here

PR #137, opened 2020-01-14
Fixes #28750 - Template controller name must be plural

ofedoren:
This will also require changes in hammer-cli-foreman-templates (update resource `:template` to `:templates`).


xprazak2:
Makes sense, test failures related though.

ofedoren:
@xprazak2, as per our discussion, I've changed the documentation only. This will do for hammer related issue.

xprazak2:
Thanks!

PR #139, opened 2020-01-31
Fixes #28630 - Show filtered results of exporting

ofedoren:
ScreenShot-1580483040974

Shows information about filtered templates in both UI and API responses.

I've refactored the code, so it looks like `import` a bit.

### DIFF
--- a/app/controllers/api/v2/template_controller.rb
+++ b/app/controllers/api/v2/template_controller.rb
@@ -37,7 +37,8 @@ module Api
       param_group :taxonomies, ::Api::V2::BaseController
       def export
         @result = ForemanTemplates::TemplateExporter.new(template_export_params).export!
-        render :json => { :message => @result.to_h }, :status => @result.exported ? 200 : 500
+        @result[:templates] = @result[:templates].map(&:to_h)
+        render :json => { :message => @result }, :status => @result[:error] ? 500 : 200
       end
     end
   end
--- a/app/services/foreman_templates/template_exporter.rb
+++ b/app/services/foreman_templates/template_exporter.rb
@@ -4,27 +4,29 @@ module ForemanTemplates
       super + %i(metadata_export_mode)
     end
 
+    def initialize(args = {})
+      super args
+      @result_lines = []
+    end
+
     def export!
-      @export_result = ExportResult.new(@repo, @branch, foreman_git_user)
       if git_repo?
         export_to_git
       else
         export_to_files
       end
-
-      return @export_result
+      export_result
     end
 
     def export_to_files
       @dir = get_absolute_repo_path
       verify_path!(@dir)
       dump_files!
-      @export_result.exported = true
     end
 
     def export_to_git
       @dir = Dir.mktmpdir
-      return @export_result if branch_missing?
+      return if branch_missing?
 
       git_repo = Git.clone(@repo, @dir)
       logger.debug "cloned '#{@repo}' to '#{@dir}'"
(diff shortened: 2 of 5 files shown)


app/controllers/api/v2/template_controller.rb
```diff
@@ -37,7 +37,14 @@ def import
       param_group :taxonomies, ::Api::V2::BaseController
       def export
         @result = ForemanTemplates::TemplateExporter.new(template_export_params).export!
-        render :json => { :message => @result.to_h }, :status => @result.exported ? 200 : 500
+        render :json => { :message => { :error => @result[:error],
```
xprazak2: Would it make sense to have a custom object that would take care of constructing the response shape so that we do not have this in controller?

app/services/foreman_templates/template_exporter.rb
```diff
@@ -42,15 +44,15 @@ def export_to_git
       if new_repo || status.added.any? || status.changed.any? || status.deleted.any? || status.untracked.any?
         git_repo.commit "Templates export made by Foreman user #{foreman_git_user}"
         git_repo.push 'origin', branch
-        @export_result.exported = true
       else
-        @export_result.warning = 'No change detected, skipping the commit and push'
+        @warning = 'No change detected, skipping the commit and push'
+        @result_lines.each { |r| r.generic_info(@warning) }
```
xprazak2: I wonder if we need to add the same message to each template when it is caused by the export as a whole.

app/services/foreman_templates/template_exporter.rb
```diff
@@ -42,15 +44,15 @@ def export_to_git
...
       end
     rescue StandardError => e
-      @export_result.error = e.message
+      @error = e.message
+      @result_lines.each { |r| r.generic_error(e.message) }
```
xprazak2: Same as the previous comment.

app/services/foreman_templates/template_exporter.rb
```diff
@@ -96,18 +94,24 @@ def get_dump_dir(template)
...
-      base = find_templates
-      if filter.present?
-        method = negate ? :reject : :select
-        base.public_send(method) { |template| template.name.match(/#{filter}/i) }
-      else
-        base
+      find_templates.each do |template|
+        if filter.present?
+          matches = template.name.match(/#{filter}/i)
+          exportable = matches ? true && !negate : false || negate
```
xprazak2: If I changed it to `matches ? !negate : negate` would I get the same result?

xprazak2:
I like the idea of having similar `ImportResult` and `ExportResult` for each template. If you add something that contains all export results + additional data about export (user, repo, error,... esentially what ExportResult did until now), export action in api controller could be really slim.

app/services/foreman_templates/template_exporter.rb
ofedoren: I think there is no reason for this currently, removed.

app/services/foreman_templates/template_exporter.rb
ofedoren: Yep, this is the same, thanks.

app/controllers/api/v2/template_controller.rb
ofedoren: @xprazak2, I'm not sure if there is a need in specific object, which will require more refactoring if we want to make it as general as it allows to reuse within `import` as well as in UI/API responses for each.

ofedoren:
@xprazak2, updated. I've removed some changes as you suggested.

xprazak2:
I like the changes, but rubocop is unhappy:

RuboCop failed!
Running RuboCop...
Inspecting 24 files
.............C..........

Offenses:

/home/jenkins/workspace/foreman_templates-pull-request/database/postgresql/label/fast/ruby/2.6/plugin/app/services/foreman_templates/template_exporter.rb:97:24: C: Performance/RedundantMatch: Use =~ in places where the MatchData returned by #match will not be used.
          exportable = template.name.match(/#{filter}/i) ? !negate : negate
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

xprazak2:
Awesome work, thanks!

PR #140, opened 2020-01-31
Fixes #23380 - Custom commit message for export

ofedoren:


xprazak2:
Nice. It would be also useful to modify the message directly in the form for individual exports. Adding the new setting into a list of [export settings](https://github.com/theforeman/foreman_templates/blob/master/app/models/setting/template_sync.rb#L13) should be enough for form field to render. You may also need to check the strong params to make sure it is whitelisted.

ofedoren:
@xprazak2, thanks! Done. Also, since `:commit_msg` is now permitted param, I've added new param docs in API controller.

xprazak2:
Looks good, needs a small change:

RuboCop failed!
Running RuboCop...
Inspecting 24 files
.C......................

Offenses:

/home/jenkins/workspace/foreman_templates-pull-request/database/postgresql/label/fast/ruby/2.6/plugin/app/controllers/concerns/foreman/controller/parameters/template_params.rb:18:13: C: Style/SymbolArray: Use %i or %I for an array of symbols.
            [:metadata_export_mode, :commit_msg]
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

xprazak2:
Thanks!

PR #149, opened 2020-07-29
Fixes #30515 - Do not mutate values object onSubmit

xprazak2:
This was not reproducible in development, only downstream:

template-import


xprazak2:
@sharvit, could you take a look please?

ares:
Thanks @xprazak2, merging!

PR #151, opened 2021-03-03
Fixes #31985 - show sync error in UI

ezr-ondrej:
The new Formik forms don't show the validation errors properly.
This should be fixed properly, but for now we are using temporary fix
introduced in https://github.com/theforeman/foreman/commit/0ed6bc5e1d8de01909293c50e62057ae22f16d1e (https://github.com/theforeman/foreman/pull/8309)

### DIFF
--- a/app/controllers/ui_template_syncs_controller.rb
+++ b/app/controllers/ui_template_syncs_controller.rb
@@ -43,6 +43,6 @@ class UiTemplateSyncsController < ApplicationController
   end
 
   def render_errors(messages, severity = 'danger')
-    render :json => { :error => { :errors => { :base => messages }, :severity => severity } }, :status => :unprocessable_entity
+    render :json => { :error => { :errors => { :base => messages }, full_messages: messages, :severity => severity } }, :status => :unprocessable_entity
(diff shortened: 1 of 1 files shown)


xprazak2:
When there is an error in form submission, user is presented with empty state which is misleading. The toast with error is shown only if user navigates back to sync form.

If there is an error, we should stay in the form and show the toast.

ezr-ondrej:
> If there is an error, we should stay in the form and show the toast.

Agreed, do you see a way how we can fix it in a simle fix?

Proper fix should include frontend validation, but I doubt we can do it in a safe to backport fashion.

xprazak2:
The form submission relies on common infrastructure provided by Foreman core which is now broken so there is hardly anything we can do about that in a plugin. 

We cannot use frontend validation to prevent preconditions for all server-side errors as we can hardly detect whether user has permissions to export into a given repository.

I think I have voiced my concerns strongly enough, here and in other places as well, so there is no point of me going on about it. Even though this does not fix the problem, it does not make it any worse, so merging. Thanks!

PR #153, opened 2021-04-12
Fixes #32291 - Fix taxonomy props in sync form

xprazak2:
The main change is using taxonomies from context, rest is moving from class-based component to function-based one so the hooks can be used.

### DIFF
--- a/webpack/components/NewTemplateSync/components/NewTemplateSyncForm/NewTemplateSyncForm.js
+++ b/webpack/components/NewTemplateSync/components/NewTemplateSyncForm/NewTemplateSyncForm.js
@@ -1,158 +1,105 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import { compose } from 'redux';
 
 import ForemanForm from 'foremanReact/components/common/forms/ForemanForm';
-import * as Yup from 'yup';
+import { useForemanLocation, useForemanOrganization } from 'foremanReact/Root/Context/ForemanContext';
+
 import SyncSettingsFields from '../SyncSettingFields';
 import SyncTypeRadios from '../SyncTypeRadios';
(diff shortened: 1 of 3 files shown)


webpack/components/NewTemplateSync/components/NewTemplateSyncForm/NewTemplateSyncForm.js
```diff
@@ -1,158 +1,105 @@
...
+  exportUrl,
+  initialValues,
+  userPermissions
+}) => {
+  const allowedSyncType = (currentUserPermissions, radioAttrs) =>
+    currentUserPermissions[radioAttrs.permission];
+
+  const radioButtons = [
+    { label: 'Import', value: 'import', permission: 'import' },
+    { label: 'Export', value: 'export', permission: 'export' },
```
tbrisker, line 29: Not related to this pr but just something i noticed, i think the labels should be marked for translation:
```suggestion
    { label: __('Import'), value: 'import', permission: 'import' },
    { label: __('Export'), value: 'export', permission: 'export' },
```

webpack/components/NewTemplateSync/components/NewTemplateSyncForm/NewTemplateSyncForm.js
```diff
@@ -1,158 +1,105 @@
...
+    <ForemanForm
+      onSubmit={(values, actions) => {
+        const url = syncType === 'import' ? importUrl : exportUrl;
+        return submitForm({
+          url,
+          values: compose(
+            addLocParams,
+            addOrgParams
+          )(values[syncType]),
+          message: `Templates were ${syncType}ed.`,
```
tbrisker, line 75: This as well

tbrisker (approved):
Thanks @xprazak2 ! while going over the code I notices a couple of string that should probably be marked for translation but that can be done separately as it is not related to this chage.

PR #155, opened 2021-05-06
Switch to using Foreman rubocop and fix most cops

tbrisker:


xprazak2:
[test foreman_templates]

xprazak2:
The rubocop offences do not seem transient...

tbrisker:
@ezr-ondrej any idea? looks like ci runs `rake foreman_templates:rubocop` which for some reason gets a different config?

ezr-ondrej:
Yup, the task is generated by foreman and runs from foreman folder, so the config is foreman config and joins it with config found on the checked path, so with the plugin config. Also, because there is include pattern defined in the rake task, you can't decide what files to check or not in the `.rubocop.yml`.
Idea: lets get rid of the rake task, it is not usefull at all. That's what we have the rubocop gem for (altough I have to admit there is still work to be done on it that is on me :()

tbrisker:
hmm, i wonder if we could run rubocop as a gh action and only trigger jenkins from the action if it passes instead of immediately? @ekohl maybe you know if that's possible?

ekohl:
In theory we can do it (you can make Jenkins listen to events when checks complete) but I don't think there's support for that in the current code.

ezr-ondrej:
I think doing it on jenkins (as well as on GH actions) is ok path to do this, it just should be in pipeline and run as first step in said pipeline. And it should be running only clone this single repo, bundle rubocop, run `bundle exec rubocop` command.

ekohl:
(testing that theory now)

tbrisker:
Looks like it didn't work:
`NoMethodError: undefined method `config=' for #<RuboCop::RakeTask:0x0000000011c26918>`
also, for some reason this plugin runs rubocop after tests instead of before??

ekohl:
> Looks like it didn't work:
> `NoMethodError: undefined method `config=' for #RuboCop::RakeTask:0x0000000011c26918`

Ah, I think we'll need to pass `task.options = ['--config', ForemanTemplates::Engine.root.join('.rubocop.yml').to_s]`. Trying that

>  also, for some reason this plugin runs rubocop after tests instead of before??

I think enhance adds it after unit tests here:
https://github.com/theforeman/foreman_templates/blob/d5e8e5b17387a5f0cc7e246f89927c86c546c581/lib/tasks/foreman_templates_tasks.rake#L104

I never really went ahead with fixing this since in https://community.theforeman.org/t/rake-task-to-run-rubocop-on-a-plugin/19979/4 it was unanimous that a central solution was not desired. If we do want to keep such a thing, my idea was to add a `rubocop:plugin[myplugin]` task that would have this. It could be run in CI before the actual unit tests.

tbrisker:
I wonder why this plugin test take so long to complete, but that's out of scope for this PR.

xprazak2:
Thanks everyone!

PR #156, opened 2021-11-21
Prepare for SettingDSL

ezr-ondrej:
we use setting class quite a lot, first I want to move the heavy lifting to `SettingRegistry`.
Then we can move to implementing the setting definitions in the DSL

### DIFF
--- a/app/views/template_sync_settings/show.json.rabl
+++ b/app/views/template_sync_settings/show.json.rabl
@@ -1,11 +1,11 @@
 object @setting
 
 node do |setting|
-  { :name => setting.short_name }
+  { :name => setting.name.delete_prefix('template_sync_') }
 end
 
 attributes :id, :value, :description, :settings_type, :default, :full_name
 
 node do |setting|
-  { :selection => (setting.select_collection&.map { |key, label| { value: key, label: label } } || []) }
+  { :selection => (setting.select_values || {}).map { |key, label| { value: key, label: label } } }
 end
--- a/app/views/template_sync_settings/show.json.rabl
+++ b/app/views/template_sync_settings/show.json.rabl
@@ -7,5 +7,5 @@ end
 attributes :id, :value, :description, :settings_type, :default, :full_name
 
 node do |setting|
-  { :selection => setting.selection }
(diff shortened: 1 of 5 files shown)


lib/foreman_templates.rb
```diff
@@ -1,4 +1,8 @@
 require 'foreman_templates/engine'
 
 module ForemanTemplates
+  BASE_SETTING_NAMES = %w(repo branch dirname filter negate)
+  COMMON_SETTING_NAMES = BASE_SETTING_NAMES | %w(verbose)
```
adamruzicka, line 5: Where is this used?

app/views/template_sync_settings/show.json.rabl
```diff
@@ -1,11 +1,11 @@
...
 node do |setting|
-  { :name => setting.short_name }
+  { :name => setting.name.delete_prefix('template_sync_') }
 end
 
 attributes :id, :value, :description, :settings_type, :default, :full_name
 
 node do |setting|
-  { :selection => setting.selection }
+  { :selection => (setting.select_values&.map { |key, label| { value: key, label: label } } || []) }
```
adamruzicka, line 10: In general I prefer providing default values as soon as possible, to me it is easier to read.
```
(setting.select_values || []).map { |k, v| ... }
```

As this is a style-only change, feel free to ignore this comment

adamruzicka (commented):
Also I'm getting this when opening the form. It doesn't seem to happen with master

image

lib/foreman_templates.rb
ezr-ondrej: nowhere, I just felt to have the common_settings full list. but I can do without.

ezr-ondrej:
The error was fixed in `master`, I only had an old base for this PR. It should work now.

adamruzicka:
In general +1, but test failures are most likely related

ezr-ondrej:
Indead, we need to reload values in the SettingRegistry to reflect the changes.

ezr-ondrej:
:green_apple:

adamruzicka (approved):
LGTM, but I don't have merge permissions

PR #159, opened 2022-01-18
Fixes #34275 - consume PF4 pagination from core

amirfefer:


webpack/components/TemplateSyncResult/components/SyncResultList.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { ListView } from 'patternfly-react';
 import PropTypes from 'prop-types';
 
-import Pagination from 'foremanReact/components/Pagination/PaginationWrapper';
+import Pagination from 'foremanReact/components/Pagination';
```
Ron-Lavi, line 5: can you also update the components new props?
```jsx
<Pagination
      itemCount={templates.length}
      {...pagination} // not sure if needed
      onChange={pageChange}
    />
```

Ron-Lavi (commented):
Thanks @amirfefer!

Ron-Lavi (commented):
Thanks @amirfefer!

ezr-ondrej (commented):
I believe we are still using the old props, we need to update those to new ones as @Ron-Lavi have mentioned.

webpack/components/TemplateSyncResult/components/SyncResultList.js
amirfefer: updated

webpack/components/TemplateSyncResult/components/SyncResultList.js
```diff
@@ -19,11 +19,9 @@ const SyncResultList = ({ pagination, pageChange, templates, editPaths }) => (
       />
     ))}
     <Pagination
-      viewType="list"
       itemCount={templates.length}
       pagination={pagination}
```
ezr-ondrej, line 23: I believe this option is unsupported we now have directly perPage, perPageOptions and so...

ezr-ondrej (commented):
I believe we will need to update snapshots and we still are passing the `pagination` property, that is no longer supported.

Ron-Lavi:
@amirfefer can you update the PR? thanks 🙏🏼

webpack/components/TemplateSyncResult/components/SyncResultList.js
amirfefer: 👍  extracted the props from it

ezr-ondrej (changes requested):
I belive we have to also change:

https://github.com/theforeman/foreman_templates/blob/a53f715d7d7f37fbaa6cc29205bf5438862aa2b9/webpack/components/TemplateSyncResult/TemplateSyncResultReducer.js#L27

Because the new onChange is sending `per_page` instead of `perPage`, so we will have to camelize it in the Reducer to get the desired state.

webpack/components/TemplateSyncResult/components/SyncResultList.js
```diff
@@ -2,13 +2,19 @@ import React from 'react';
...
+import Pagination from 'foremanReact/components/Pagination';
 
 import SyncedTemplate from './SyncedTemplate';
 import { templatesPage } from '../TemplateSyncResultHelpers';
 import ListViewHeader from './ListViewHeader';
 
-const SyncResultList = ({ pagination, pageChange, templates, editPaths }) => (
+const SyncResultList = ({
+  pagination,
+  pagination: { page, perPage },
```
ezr-ondrej, line 13: Well I thought we could just camalize in the reducer, but now this will be wrong, won't it?
This should now be
```suggestion
  pagination: { page, perPage: per_page },
```

amirfefer: tested, should work now :)
at first, I tried to camel case the pagination object, but it ended with some errors 🤔 

Also, I created https://github.com/theforeman/foreman/pull/9097 to fix the missing variant, so the page selection dropdown  will be open correctly

ezr-ondrej:
I think we are not running JS tests here at all, so I do not expect failures, but the snapshot should be updated IMHO

amirfefer:
snapshots updated
also open https://github.com/theforeman/foreman_templates/pull/160 for adding github actions for lint / test

ezr-ondrej (commented):
I'm sorry to nitpick, but could we update the fixtures and then update snapshots?
https://github.com/theforeman/foreman_templates/blob/a53f715d7d7f37fbaa6cc29205bf5438862aa2b9/webpack/components/TemplateSyncResult/components/__tests__/SyncResultList.test.js#L21
The snapshot seems bit useless without the `per_page` option.

Apart of it LGTM :+1:

amirfefer:
no worries :)
done

ezr-ondrej (approved):
I'm good here, thanks @amirfefer ! :+1:
@theforeman/foreman_templates can we have someone to approve this (ideally release a new version for foreman 3.2)

ofedoren:
Thanks, @amirfefer! Could you please rebase just to see if JS tests are running?

ofedoren (approved):
Nevermind, @amirfefer, tests are passing locally and the changes work as expected :)

PR #167, opened 2023-01-26
Fixes #36016 - Migration error 'column settings.category does not exist'

stejskalleos:
See:

https://github.com/theforeman/foreman/pull/9524
https://github.com/theforeman/foreman/pull/9524#issuecomment-1404661119


ofedoren (approved):
Thanks, @stejskalleos, let's get this in.

UPD: I'll update the sommit message to match the issue before merging.

PR #168, opened 2023-01-26
Fixes #36009 - Fix untranslatable strings

MariaAga:
String extraction relies on static strings.


ofedoren (approved):
Thanks, @MariaAga, makes more sense than try to translate "dynamic" strings.

PR #171, opened 2023-05-08
Bump theforeman/builder to 12+

ehelms:
The current version of foreman_templates will not build in our environment without this change.


ofedoren (commented):
Thanks, @ehelms, but could we make `@theforeman/*` deps more open since we're at it? There are plugins with more permissive deps:

https://github.com/theforeman/foreman_ansible/blob/master/package.json

https://github.com/theforeman/foreman_webhooks/blob/master/package.json

package.json
```diff
@@ -25,7 +25,7 @@
   },
   "devDependencies": {
     "@babel/core": "^7.7.0",
-    "@theforeman/builder": "^10.1.0",
+    "@theforeman/builder": ">= 12.0.1",
```
ofedoren, line 28: Thanks, @ehelms, but what I meant is to update all others as well (I think they should be synced?..)

```suggestion
    "@theforeman/builder": ">= 12.0.1",
    "@theforeman/eslint-plugin-foreman": ">= 12.0.1",
    "@theforeman/vendor-dev": ">= 12.0.1",
    "@theforeman/test": ">= 12.0.1",
    "@theforeman/stories": ">= 12.0.1",
```

ehelms: Done.

ofedoren (approved):
Thanks, @ehelms!

PR #174, opened 2023-05-25
Add packit config

ofedoren:


ofedoren:
Weird... It seems the error for packit if `No matching package to install: 'npm(@theforeman/builder) < 11.0.0'`, but we updated `@theforeman/builder` version dep to `>=12.0.1`...

ofedoren:
That should be solved by https://github.com/theforeman/foreman-packaging/pull/9431.

PR #177, opened 2023-11-01
i18n - pulling from tx

adamruzicka:


ofedoren (approved):
Thanks, @adamruzicka!

PR #179, opened 2023-11-28
Sync locale Makefile from plugin template

ekohl:


ofedoren (approved):
Thanks, @ekohl!

packit failure is regarding `node-sass`, so not related.

PR #180, opened 2024-01-12
Add shared GitHub Actions workflow for ruby

ofedoren:


ekohl (commented):
There's no rubocop task in the Rakefile:
```
Run if [ -f Rakefile ] ; then
rake aborted!
Don't know how to build task 'rubocop' (See the list of available tasks with `rake --tasks`)
```

I raised it before, but should we start assuming Rubocop as an executable exists and always use that?

ekohl:
https://github.com/theforeman/foreman_webhooks/pull/71/files#r1450828173 is also an option.

.github/workflows/ruby_ci.yml
```diff
@@ -0,0 +1,19 @@
...
+on: pull_request
+
+concurrency:
+  group: ${{ github.ref_name }}-${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  rubocop:
+    name: Rubocop
+    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
```
ekohl, line 12: > I raised it before, but should we start assuming Rubocop as an executable exists and always use that?

For now I'd go with specifying the exact command.

```suggestion
    uses: theforeman/actions/.github/workflows/rubocop.yml@v0
    with:
      command: bundle exec rubocop --parallel --format github
```

ekohl:
@ofedoren looks like there are RuboCop failures. Mind taking a look?

archanaserver:
@ekohl i'll take a look and raise a PR with the rubocop fix here

archanaserver:
This should fix it now: https://github.com/theforeman/foreman_templates/pull/182

ofedoren:
I've fixed rubocop here, but the tests are now red due to webpack changes. I'm not so sure how to fix it though. 

@MariaAga, sorry for bothering you, I think you already work on that, but just to make sure: should we wait a bit or should we do something to fix the issues?

MariaAga:
Could you change css files to:
```
 - @import '~@theforeman/vendor/scss/variables';
+ @import 'node_modules/@theforeman/vendor/scss/variables.scss';
```
Its failing tests without it and its not a priority to investigate right now as its only a test issue

PR #190, opened 2024-09-20
Fixes #37850 - Support Rails 7.0

ofedoren:

PR #192, opened 2024-11-06
Fixes #37986, #37982 - Fix UI Error when importing templates

adamlazik1:
Importing templates through UI causes an error and the page with import results does not load.

Steps to Reproduce:

1. Have foreman instance with the templates plugin on nightly (error originates in one of the newest commits so it is not on stream yet).

2. Navigate to Hosts > Templates > Sync Templates

3. Import templates from https://github.com/theforeman/community-templates.git

Actual behavior:

The page with results does not load and two popups are displayed. One saying that templates have been imported succesfully and another saying that there has been an error.

Expected behavior:

Page with results loads normally and displays templates correctly.

It appears that UI expects the error messages to be wrapped in array as opposed to being just string, and hammer (https://github.com/theforeman/hammer-cli-foreman-templates/pull/15) expects an empty hash instead of nil when there are no errors.

### DIFF
--- a/app/services/foreman_templates/parse_result.rb
+++ b/app/services/foreman_templates/parse_result.rb
@@ -16,7 +16,7 @@ module ForemanTemplates
         :additional_errors => @additional_errors,
         :additional_info => @additional_info,
         :exception => @exception&.message,
-        :validation_errors => errors,
+        :validation_errors => transformed_errors,
         :file => @template_file,
         :type => @template.present? ? @template.class.name.underscore : nil
       }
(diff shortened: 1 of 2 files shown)


ofedoren (changes requested):
Thanks @adamlazik1 for taking this further. I'd propose a bit different fix:

Instead of
```ruby
:validation_errors => errors,
```
try
```ruby
:validation_errors => errors || {},
```

ofedoren:
@adamlazik1, yeah, sorry about that. It seems #190 was quite wrong, it should have been like this: https://github.com/ofedoren/foreman_templates/pull/1/files

It also reverts changes in test file to the original state.

adamlazik1:
Thanks @ofedoren, I applied your patch and it fixes both the UI and the hammer issue.

ofedoren (approved):
Thanks, @adamlazik1 !

PR #194, opened 2025-02-06
Fixes #38195 - Hide http proxy setting when syncing through ssh

adamlazik1:
Currently, the HTTP proxy policy setting in ui is displayed at all times. However it only goes into effect when syncing through HTTP or HTTPS. SSH sync does not use the proxy, therefore the user shoudn't be able to see the proxy policy if a protocol different from HTTP(S) is specified in the repo url.

### DIFF
--- a/app/controllers/api/v2/template_controller.rb
+++ b/app/controllers/api/v2/template_controller.rb
@@ -13,7 +13,8 @@ module Api
         param :filter, String, :required => false, :desc => N_("Export templates with names matching this regex (case-insensitive; snippets are not filtered).")
         param :negate, :bool, :required => false, :desc => N_("Negate the prefix (for purging).")
         param :dirname, String, :required => false, :desc => N_("Directory within Git repo containing the templates.")
-        param :http_proxy_policy, ForemanTemplates.http_proxy_policy_types.keys, :required => false, :desc => N_("HTTP proxy policy for template sync. If you choose 'selected', provide the `http_proxy_id` parameter.")
+        param :http_proxy_policy, ForemanTemplates.http_proxy_policy_types.keys, :required => false, :desc => N_("HTTP proxy policy for template sync. \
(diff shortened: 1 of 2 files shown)


adamruzicka (commented):
Could we also mention it in apidoc?

Edit: The ui part seems to work well

adamlazik1:
Mention in apidoc added.

app/controllers/api/v2/template_controller.rb
```diff
@@ -13,7 +13,8 @@ class TemplateController < ::Api::V2::BaseController
         param :filter, String, :required => false, :desc => N_("Export templates with names matching this regex (case-insensitive; snippets are not filtered).")
         param :negate, :bool, :required => false, :desc => N_("Negate the prefix (for purging).")
         param :dirname, String, :required => false, :desc => N_("Directory within Git repo containing the templates.")
-        param :http_proxy_policy, ForemanTemplates.http_proxy_policy_types.keys, :required => false, :desc => N_("HTTP proxy policy for template sync. If you choose 'selected', provide the `http_proxy_id` parameter.")
+        param :http_proxy_policy, ForemanTemplates.http_proxy_policy_types.keys, :required => false, :desc => N_("HTTP proxy policy for template sync. \
+          This setting is only effective when synchronizing templates through the HTTP or the HTTPS protocol. If you choose 'selected', provide the `http_proxy_id` parameter.")
```
aneta-petrova, line 17: I suppose that if you really don't want users to use this setting in other situations, you can be more direct with the wording:
```suggestion
          Use only when synchronizing templates through the HTTP or the HTTPS protocol. If you choose 'selected', provide the `http_proxy_id` parameter.")
```

adamlazik1: Thanks, applied.

adamruzicka:
Thank you @adamlazik1 & @aneta-petrova !

PR #200, opened 2025-04-02
Fixes #38195 - Hide HTTP proxy selection if url is not HTTP(S)

adamlazik1:
Currently, there is an issue in UI template sync that if you first enter an HTTP url, then you select custom HTTP proxy, and then you change the url to non-HTTP(S) form, the HTTP proxy field remains. The field should be only visible if the url in the in HTTP(S) form, so I added a simple check into the condition which prevents this bug from occuring.


adamruzicka (approved):
Lgtm

PR #203, opened 2025-07-01
Fixes #37907 - move css import from vendor to foreman

MariaAga:
css imports were moved in https://github.com/theforeman/foreman/pull/10345 already
and will break after https://github.com/theforeman/foreman/pull/10342 is merged.
This pr is only for the css import and can be merged now


adamruzicka (approved):
I took it for a spin and haven't seen anything out of place, let's get it in

